Issue #21, opened 2021-06-02
Unable to update a filter field of a saved filter


0xedward:
Thanks for your continued interest in helping us out with Pysa and SAPP, @m0mosenpai and @gracewgao!

## Scenario
Currently, we don't support updating existing SAPP filters. If a user wants to change a field for an existing filter, but keep the name of the filter the same, they would need to delete the filter before re-saving the filter with the same name

For example, suppose we have an existing filter `Test Filter 1` that filters only for `Code==5008` and we want to overwrite `Test Filter 1` to only filter for `Code==5003`. We expect to perform this action by loading the saved filter `Test Filter 1`, changing the value of `code` from `5008` to `5003`, and finally saving the change to `Test Filter 1`. However to perform this action now, we will need to delete `Test Filter 1`, set the `code` to filter for `5003`, and save a new filter with the same name `Test Filter 1`

## Steps to Reproduce
1. Set up your dev virtual environment
```
git clone git@github.com:facebook/sapp.git && cd sapp
python3 -m venv ~/.venvs/sapp
source ~/.venvs/sapp/bin/activate
pip3 install -r requirements.txt
cd sapp/ui/frontend && npm install
npm run-script build
```
2. Run Pysa on a repo and store the results to pass into SAPP to ingest
```
pyre analyze --no-verify --save-results-to .
```
3. Use SAPP to ingest the results from the Pysa run
```
python3 -m sapp.cli analyze taint-output.json
```
4. Run SAPP server and go to http://localhost:5000
```
python3 -m sapp.cli server
```
5. Create a filter with for any field (eg. filter for code `5008`) and save the filter as `Test Filter 1`
![image](https://user-images.githubusercontent.com/14011954/120400238-7e8a9500-c30b-11eb-9721-b8b949c8a4c4.png)
6. Fill some different filter fields from the ones you have set in step 5 and save the filter with the same name as in step 5 (e.g. `Test Filter 1`)
7. Notice that the filter fails to overwrite the previously saved filter and you get the following traceback in SAPP server logs
```
2021-06-01 19:03:22,221 [ERROR] Traceback (most recent call last):
...
graphql.error.located_error.GraphQLLocatedError: (sqlite3.IntegrityError) UNIQUE constraint failed: filters.name
[SQL: INSERT INTO filters (name, description, json) VALUES (?, ?, ?)]
[parameters: ('test', None, '{"features":[{"mode":"all of","features":[]}],"codes":[5001]}')]
[...]

PR #22, opened 2021-06-07
Added updating to saved filters

gracewgao:
This PR refers to #21 and adds the ability to update the filter field of a saved field from the web UI. 

To do this, there’s just a couple of new changes in `filters.py` which inserts the new filter into the database or updates it (if the filter name already exists).

I first dug around a bit in the docs to see if there was anything like `UPSERT` available for SQLAlchemy, but I think many people online recommended to just query the database and update the fields manually (which seems to be alright, since there are only two fields that need to be updated in the object). I manually tested updating the filter fields manually to check that it works for all the filter options (codes, paths, callables, features, status, and trace lengths).

Please let me know if I missed something or if there’s anything that should be fixed/improved!

### DIFF
--- a/sapp/ui/filters.py
+++ b/sapp/ui/filters.py
@@ -87,7 +87,9 @@ def save_filter(session: Session, filter: Filter) -> None:
         session.add(FilterRecord.from_filter(filter))
         LOG.debug(f"Adding {filter}")
     else:
+        # pyre-ignore [8]: `graphene.String` is not compatible with `str`
         existing.description = filter.description
+        # pyre-ignore [8]: `graphene.String` is not compatible with `str`
         existing.json = filter.json
         LOG.debug(f"Updating {filter}")
 
--- a/sapp/ui/filters.py
+++ b/sapp/ui/filters.py
@@ -79,8 +79,10 @@ def all_filters(session: Session) -> List[Filter]:
 
 def save_filter(session: Session, filter: Filter) -> None:
 
-    existing = session.query(FilterRecord).filter(FilterRecord.name == filter.name).first()
-    
+    existing = (
+        session.query(FilterRecord).filter(FilterRecord.name == filter.name).first()
+    )
+
     if not existing:
         session.add(FilterRecord.from_filter(filter))
         LOG.debug(f"Adding {filter}")
--- a/sapp/ui/filters.py
+++ b/sapp/ui/filters.py
@@ -78,8 +78,17 @@ def all_filters(session: Session) -> List[Filter]:
 
 
 def save_filter(session: Session, filter: Filter) -> None:
-    LOG.debug(f"Storing {filter}")
-    session.add(FilterRecord.from_filter(filter))
+
+    existing = session.query(FilterRecord).filter(FilterRecord.name == filter.name).first()
+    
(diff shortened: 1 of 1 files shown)


sapp/ui/filters.py
```diff
@@ -78,8 +78,17 @@ def all_filters(session: Session) -> List[Filter]:
 
 
 def save_filter(session: Session, filter: Filter) -> None:
-    LOG.debug(f"Storing {filter}")
-    session.add(FilterRecord.from_filter(filter))
+
+    existing = session.query(FilterRecord).filter(FilterRecord.name == filter.name).first()
```
arthaud, line 82: Our linter is unhappy and wants:
```
existing = (
    session.query(FilterRecord).filter(FilterRecord.name == filter.name).first()
)
```

arthaud, line 82: Our linter is unhappy and wants:
```
existing = (
    session.query(FilterRecord).filter(FilterRecord.name == filter.name).first()
)
```

arthaud, line 82: Our linter is unhappy and wants:
```python
existing = (
    session.query(FilterRecord).filter(FilterRecord.name == filter.name).first()
)
```

sapp/ui/filters.py
```diff
@@ -78,8 +78,17 @@ def all_filters(session: Session) -> List[Filter]:
...
+    
+    if not existing:
+        session.add(FilterRecord.from_filter(filter))
+        LOG.debug(f"Adding {filter}")
+    else:
+        existing.description = filter.description
+        existing.json = filter.json
```
arthaud, line 89: don't we need to call `existing.save()`?

arthaud (changes requested):

sapp/ui/filters.py
gracewgao: I think it saves by default!

arthaud: My bad, I missed the `session.commit()` below.

arthaud, line 89: Could you add `#pyre-ignore` to both of these? The type checker claims that the left hand side is `Optional[str]` but the right hand side is `graphene.String`. I'm not sure why but it seems like `graphene.String` doesn't extend `str`?

```python
# pyre-ignore [8]: `graphene.String` is not compatible with `str`
existing.description = filter.description
# pyre-ignore [8]: `graphene.String` is not compatible with `str`
existing.json = filter.json
```

gracewgao:
Hey, just wanted to drop a comment and let you know that I committed the suggestions regarding the linting and pyre-ignore issues, but let me know if there's anything else!

gracewgao:
@gbleaney Also, just as an update - I think last week we talked about finding a better way to upsert and there was the suggestion of `on_conflict_do_update` and other alternatives mentioned [on StackOverflow](https://stackoverflow.com/a/44395983), but I had a lot of errors when I tried to implement it (I'm not sure if we're using Postgres?). I'm just going to leave it the way it is for now unless you think this is worth pursuing more?

0xedward:
Sorry about the delay on merging this! Thanks for working on SAPP and making it better, @gracewgao! 😄

PR #25, opened 2021-07-13
Adds ability to copy source code in a Trace

abishekvashok:
Add ability to copy source code in a trace. Previously copy, cut, and
paste operations were disabled. Now enabled copy. Even though paste and
cut appears in browser ui, nothing happens when they are clicked.

Fixes https://github.com/MLH-Fellowship/pyre-check/issues/28

### DIFF
--- a/sapp/ui/frontend/src/Source.css
+++ b/sapp/ui/frontend/src/Source.css
@@ -51,3 +51,7 @@
 ::-webkit-resizer {
   border-bottom: 2px solid lightgray;
 }
+
+.CodeMirror-cursor {
+  display: none !important;
+}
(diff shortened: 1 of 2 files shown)


abishekvashok:
Before: 
Screenshot 2021-07-13 at 8 50 46 PM

After:
Screenshot 2021-07-13 at 8 51 16 PM

r0rshark:
Curious what's the difference with `readOnly: true` https://github.com/codemirror/CodeMirror/issues/1099 ?

abishekvashok:
@r0rshark seems like, it's better, the cut and paste operations seems disabled, the problem was the blinking cursor, but I added css code to fix that!

Screenshot 2021-07-14 at 3 38 48 PM

Thanks for linking the issue :D

PR #26, opened 2021-07-14
Allow users to delete previously imported runs

abishekvashok:
Allow users to delete previously imported runs through the Web UI.
Modifies Run.js to include new delete icon. Adds new mutation to
schema.py which calls delete run function through run.py.

The delete run function deletes instances of a run with the given id,
issue instances that bind an issue with a run, and also tuples from
others tables that reference run_id.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/27

### DIFF
--- a/sapp/ui/frontend/src/Runs.js
+++ b/sapp/ui/frontend/src/Runs.js
@@ -8,10 +8,10 @@
  * @flow
  */
 
-import React from 'react';
-import {useQuery, gql} from '@apollo/client';
+import React, {useState} from 'react';
+import {useQuery, useMutation, gql} from '@apollo/client';
 import {Breadcrumb, Card, Col, Modal, Row, Typography} from 'antd';
-import {LoadingOutlined, SyncOutlined} from '@ant-design/icons';
+import {DeleteOutlined, LoadingOutlined, SyncOutlined} from '@ant-design/icons';
 
 const {Text, Link} = Typography;
 
@@ -22,6 +22,7 @@ type RunDescription = $ReadOnly<{
 
 function Run(props: $ReadOnly<{run: RunDescription}>): React$Node {
   const gutter = [8, 8];
+  const [showRun, setShowRun] = useState(true);
 
   const Label = (props: $ReadOnly<{children: React$Node}>): React$Node => {
     return (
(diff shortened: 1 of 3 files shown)


abishekvashok:
Working Screenshots:

Screenshot 2021-07-14 at 9 10 08 PM
Screenshot 2021-07-14 at 9 10 11 PM

sapp/ui/frontend/src/Runs.js
```diff
@@ -38,7 +39,29 @@ function Run(props: $ReadOnly<{run: RunDescription}>): React$Node {
...
+
+  const onDelete = (): void => {
+    deleteRun({variables: {id: props.run.run_id}});
+    setShowRun(false);
+  };
+
+  if(deleteError) {
+    Modal.error({title: 'Unable to delete run ', content: deleteError.toString()});
+    return null;
+  }
```
arthaud, line 62: Disclaimer: react noob here!
Shouldn't this be inside the `onDelete` function? How could `deleteError` have a value before it is actually called by `onDelete`?

arthaud (changes requested):

sapp/ui/frontend/src/Runs.js
abishekvashok: Hey, sorry for the late reply. I am not that familiar in React myself as well, so even though I knew how it works, I wanted to be sure. deleteError is derived from useMutation() and is a react hook [https://www.apollographql.com/docs/tutorial/mutations/#apply-the-usemutation-hook] 

So hooks alter the flow of the program when their value changes in a nutshell, so when you change its value all control instructions that use it are re-evaluated. Hence, we get the error as per the screenshot :) I reproduced this by modifying the graphene query to some illogical mutation name...

Screenshot 2021-07-16 at 4 17 50 PM

[Ref to hooks: https://reactjs.org/docs/hooks-intro.html]

abishekvashok: But yeah, thanks for this comment, I did a hard testing of the error message and found there was a slight bug that the run disappeared (to appear on reload) even incase of error. Fixed that in the latest change. So, thank you!

r0rshark:
Great! thanks :) The "run" Card look a bit weird since if basically takes all the width of the screen but it's mostly empty. I am not a great designer but what do you think about making them smaller and split in columns like https://ant.design/components/card/#components-card-demo-in-column ?

abishekvashok:
@r0rshark done, the suggestion was great - it looks pretty now:
Screenshot 2021-07-19 at 5 37 55 PM

PR #27, opened 2021-07-19
Create easier access to selecting saved filters

abishekvashok:
Creates easier access to selecting web filters by moving saved filters
out of the filter form to a seperate popup. Now clicking the saved
filter applies the filter (and perfomes the page reload).

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/30

### DIFF
--- a/sapp/ui/frontend/src/Filter.css
+++ b/sapp/ui/frontend/src/Filter.css
@@ -33,3 +33,11 @@
 .ant-collapse-content-box{
     padding: 1px !important;
 }
+
+.ant-layout-sider{
+    background: none !important;
+}
+
+.ant-menu-submenu-title{
+    padding-left: 16px !important;
+}
--- a/sapp/ui/frontend/src/Filter.js
+++ b/sapp/ui/frontend/src/Filter.js
@@ -16,11 +16,11 @@ import {
   Checkbox,
   Collapse,
   Divider,
-  Dropdown,
   Popover,
   Button,
   Form,
   Input,
+  Layout,
   Menu,
   Modal,
   Row,
@@ -31,10 +31,11 @@ import {
   Typography,
 } from 'antd';
 import {
+  FieldTimeOutlined,
   FilterOutlined,
   PlusOutlined,
   MinusCircleOutlined,
-  MoreOutlined,
+  SettingOutlined,
   SaveOutlined,
   DeleteOutlined,
 } from '@ant-design/icons';
@@ -45,6 +46,8 @@ import './Filter.css';
 
 const {Option} = Select;
 const {Panel} = Collapse;
+const {Sider} = Layout;
+const {SubMenu} = Menu;
 const {Text} = Typography;
 
 type FeatureCondition = {
@@ -651,13 +654,11 @@ const FilterForm = (props: {
   setVisible: boolean => void,
   currentFilter: FilterDescription,
   setCurrentFilter: FilterDescription => void,
+  appliedFilter: FilterDescription,
+  setAppliedFilter: FilterDescription => void,
 }): React$Node => {
-  const [appliedFilter, setAppliedFilter] = useState<FilterDescription>(
-    loadFilter() || emptyFilter,
-  );
-
   const apply = () => {
-    setAppliedFilter(props.currentFilter);
+    props.setAppliedFilter(props.currentFilter);
     props.refetch(filterToVariables(props.currentFilter));
     props.setVisible(false);
   };
@@ -736,7 +737,7 @@ const FilterForm = (props: {
           type="primary"
           onClick={apply}
           loading={props.refetching}
-          disabled={filterEqual(props.currentFilter, appliedFilter)}>
+          disabled={filterEqual(props.currentFilter, props.appliedFilter)}>
           Apply
         </Button>
       </div>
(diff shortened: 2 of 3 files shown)


abishekvashok:
Screenshots:
Screenshot 2021-07-19 at 5 21 14 PM

r0rshark:
I am not totally convinced about this UI. It's a bit confusing the we now have 2 different forms for the filter and they can overlap like in this picture:
Screenshot 2021-07-21 at 10 10 39
Moreover selecting a saved filter would still require 3 clicks instead of 4 which is not a great improvement.

I can imagine 2 approaches:
1. Keep the Filters like it is now and once clicking on a Saved Filter automatically apply it. This would reduce the interaction from 4 -> 3
2. Have a sidebar which shows the filter and allows to quickly select them. Interaction would be just one click (internally we have something similar)
Screenshot 2021-07-21 at 10 21 31

0xedward:
cc @dkgi if he has opinions on the UI design here

gbleaney:
+1 to @r0rshark's sidebar idea

dkgi:
Personally think what we have internally is pretty terrible and it be great if we could figure out how to do better here that'd be great. Maybe this could be a side-panel with three sections:

1. "Create": what we currently have hidden in the button and would be collapsed by default
2. "Search": the search for existing ones as shown in this diff
3. "Recent": a list of filters the user has looked up recently (maybe stored in the browser storage?)

abishekvashok:
@r0rshark @gbleaney @0xedward @dkgi 
Here you go, sidebar as everyone suggested :)

https://user-images.githubusercontent.com/8947010/126631543-37aaa2f5-32ed-4283-b1e0-a165cda15383.mov

abishekvashok:
The filter wasn't removed from recent filters on delete, sorry I messed that up, but fixed now :)

r0rshark:
@dkgi what do you think about this approach? Is it close enough to your ideas? :)

dkgi:
@abishekvashok this looks pretty awesome! Couple of things I notice:

1. I think the current version might be a bit confusing to users since the users attention has to move across the full screen when selecting a filter to apply it. I'm wondering if we could move the whole content of the filter dialog into the sidebar (or would that be to small?).
2. I think it's visually confusing that the `Runs / Run 8` and the issues don't align anymore. Maybe if we move the sidebar to the right?

abishekvashok:
@dkgi thank you for your time and valuable pointers.

yep, suggestion 2 definitely makes sense. But I am not sure if we have enough room for the entire filter form in the side panel, especially since it can introduce multiple levels of collapses. But I am not that good at UI, so I'll let you pick :)

suggestion 1 and 2:
Screenshot 2021-07-27 at 3 07 59 PM

suggestion 2 only:
Screenshot 2021-07-27 at 2 46 08 PM


Let me know which of the above looks good to you - or even something other than this  - and I'll push the commits as needed :)

dkgi:
Don't have strong opinions here. Option 2 does look a bit squeezed. I could also see the list of recent filters becoming too long and hiding the filter configuration at the bottom. Maybe we could have the filter options first, in a separate card, that is collapsible and collapsed by default? If the size is an issue we could also just move the filter button down and make it the same width as the card with the recent filters to make this visually more coherent.

Curious to hear what others think though.

abishekvashok:
> If the size is an issue we could also just move the filter button down and make it the same width as the card with the recent filters to make this visually more coherent.

Nice suggestion @dkgi ! I tried this because making filters as a collapsible card would look like the same collapsed.
How does this look?

Screenshot 2021-07-28 at 2 23 00 PM

CC: @r0rshark @0xedward @gbleaney

dkgi:
We might want to tweak this a bit in follow-ups but generally this looks reasonable to me. Maybe change "Filter" to "Filter..." to make the pop-out a bit less surprising. Curious to hear what others think.

abishekvashok:
Done
Screenshot 2021-07-30 at 12 29 12 PM

r0rshark:
The UI looks good to me but I am having trouble testing this locally
```
git fetch origin pull/27/head:easier_filters
git checkout easier_filters
npm install
npm run-script build
```
shows an Error:
```
  Line 894:48:  React Hook "useMutation" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?  react-hooks/rules-of-hooks
```
Which I think is caused by moving the deleteFilterMutation from the FilterModal to the FilterControls. @abishekvashok  can you have a look at this?

abishekvashok:
@r0rshark it was due to the position where we check if we have a filter fetching error, it should be after the mutation.

We moved the mutation and associated objects out of the model to FilterControls because it's part of the sidebar and not the model anymore :)

I actually used modal.error there and after testing I reviewed the diff to double check and saw a model and alert, so changed to alert to make everything uniform but didn't test - my mistake! Fixed it now.

r0rshark:
Maybe I am missing something but trying this 
```
git fetch origin pull/27/head:easier_filters2
git checkout easier_filters2
npm install
npm run-script build
```
the page doesn't load and the console shows:
```
react-dom.production.min.js:209 TypeError: Cannot read property 'indexOf' of null
    at SubMenu.js:591
    at new i (connect.js:52)
```
@abishekvashok Can you reproduce?

abishekvashok:
Hey, I couldn't reproduce the issue at first- but when I tried deleting the recent filters I got it, and it took me a bit of time to figure out why and it was due to setting defaultOpenKeys (conditionally) to [] when we don't have a recent filter, but yeah, figured another way todo it and it works now - sorry for back and forth but it should be good for testing.

abishekvashok:
Fixed merge conflicts :)

r0rshark:
Nice looks good to me. Only minor nit can we keep the "Recent Filter" always in the same order (maybe alphabetical) it's a bit confusing they keep changing position after having clicked on them

abishekvashok:
Hey @r0rshark done! Now it doesn't move to top. I didn't do the alphabetical one, because then the filter that is first in the list in the db when used will always remain on top! Nice suggestion btw, now UX looks cooler :)

PR #28, opened 2021-07-20
Adds clear filter button

abishekvashok:
Adds clear filter button beside the Filter button to clear spplied
filters instantly thereby reducing the number of interactions required
to stop filtering to one.

The button will only be visible if filters are applied.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/34

### DIFF
--- a/sapp/ui/frontend/src/Filter.js
+++ b/sapp/ui/frontend/src/Filter.js
@@ -38,6 +38,7 @@ import {
   SettingOutlined,
   SaveOutlined,
   DeleteOutlined,
+  ClearOutlined,
 } from '@ant-design/icons';
 import {Documentation} from './Documentation.js';
 import {statusMap} from './Issue.js';
@@ -827,6 +828,23 @@ export function loadFilter(): ?FilterDescription {
   return null;
 }
 
+export function clearFilter(
+  refetch: any = false,
+  setCurrentFilter: any = false,
+  currentFilter: FilterDescription = emptyFilter,
+): void {
+  if(!refetch) {
+    window.sessionStorage.removeItem('filter');
+  } else  {
+    setCurrentFilter(emptyFilter);
+    // workaround for https://github.com/apollographql/react-apollo/issues/3709
+    // implemented in clearAndRefetch function makes us implement this hacky
+    // solution. Otherwise, the app just keeps loading and does nothing.
+    refetch(filterToVariables(emptyFilter));
+    refetch(filterToVariables(currentFilter));
+  }
+}
+
 export function filterToVariables(filter: FilterDescription): mixed {
   const rangeValue = value => {
     if (value === 0) {
@@ -1095,6 +1113,17 @@ const Filter = (props: {
     </div>
   );
 
+  const clearButton = (
+    <Button
+      style={{width: '100%', marginTop: '10px'}}
+      icon={<ClearOutlined />}
+      onClick={() =>
+        clearFilter(props.refetch, props.setCurrentFilter, props.currentFilter)
+      }>
+        Clear Filter
+    </Button>
+  );
+
   return (
     <>
       <div style={{marginBottom: '10px'}}>
@@ -1110,6 +1139,7 @@ const Filter = (props: {
             Filter...
           </Button>
         </Popover>
+        { !filterEqual(initialFilter, emptyFilter) ? clearButton : null }
       </div>
     </>
   );


abishekvashok:
Screen recording of the changes:

https://user-images.githubusercontent.com/8947010/126330640-87ebf94b-96ae-4639-b495-006c9b8c57b5.mov

r0rshark:
Looks Good :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -956,6 +957,13 @@ const Filter = (props: {refetch: any, refetching: boolean}) => {
     window.sessionStorage.setItem('filter', JSON.stringify(currentFilter));
   }, [currentFilter]);
 
+  const clearFilter = () => {
```
0xedward, line 960: Curious why we have two `clearFilter` functions (the one here and one introduced in [PR#29](https://github.com/facebook/sapp/pull/29/files#diff-356ada66e790543c2d35ebb56b522911990d711a0f46478a7fc45f4ddb89a49fR923)). Are we unable to use `window.sessionStorage.remove('filter')` here, because it would require calling `window.location.reload()` for the UI to update with the changes?

Would there be a way to only have one `clearFilter` function we use for this PR and PR#29?

sapp/ui/frontend/src/Filter.js
```diff
@@ -956,6 +957,13 @@ const Filter = (props: {refetch: any, refetching: boolean}) => {
...
+    setCurrentFilter(emptyFilter);
+    // Hacky solution to the hacky clearAndRefetch function
+    props.refetch(filterToVariables(emptyFilter));
```
0xedward, line 963: It seems like we set `currentFilter` to `emptyFilter` then effectively call `props.refetch(filterToVariables(emptyFilter))` twice. What's the reason for calling `props.refetch` twice here? Is it because the UI doesn't update after the first `props.refetch` call?

0xedward (changes requested):
Thanks for working on this! I had a some questions in `clearFilter` :)

sapp/ui/frontend/src/Filter.js
abishekvashok: ya it doesn't, the clearAndRefetch function is done in such a way to reload only when new data is available which a single refetch call isn't being able to do!

sapp/ui/frontend/src/Filter.js
abishekvashok: One of the function deletes the filter from local session storage and the other removes all currently applied filters.

0xedward: React noob here, but would something like this work instead?
```javascript
 const clearFilter = (refreshUI: boolean = true) => {
    window.sessionStorage.removeItem('filter');
    if (refreshUI) {
      props.refetch();
      setCurrentFilter(emptyFilter);
    }
  };
```

For PR#29, we can call the same function, but pass in `refreshUI=false` instead

abishekvashok: I will try

abishekvashok:
@0xedward so I updated both the PRs to contain a common function. It works well :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -920,6 +921,21 @@ export function loadFilter(): ?FilterDescription {
   return null;
 }
 
+export function clearFilter(
+  refetch: any = false,
+  setCurrentFilter: any = false,
+  currentFilter: FilterDescrition = emptyFilter,
```
0xedward, line 927:
```suggestion
  currentFilter: FilterDescription = emptyFilter,
```

sapp/ui/frontend/src/Filter.js
```diff
@@ -920,6 +921,21 @@ export function loadFilter(): ?FilterDescription {
...
+) {
+  if(!refetch) {
+    window.sessionStorage.removeItem('filter');
+  } else  {
+    setCurrentFilter(emptyFilter);
+    // Hacky solution to the hacky clearAndRefetch function
```
0xedward, line 933: For future folks looking at this code, could you replace this comment to include why `refetch` needs to be called twice?

0xedward (changes requested):
Had some suggestions regarding typo with a type annotation and providing context with comments in code

sapp/ui/frontend/src/Filter.js
abishekvashok: Sure

abishekvashok:
Done :)

r0rshark:
@abishekvashok Just curious if you tried the @0xedward  suggestion [here](https://github.com/facebook/sapp/pull/28#discussion_r675192680)

abishekvashok:
@r0rshark yep, both the prs have a common function now!

r0rshark:
@abishekvashok  maybe I am missing something but @0xedward  suggestion [here](https://github.com/facebook/sapp/pull/28#discussion_r675192680) was to have something like:
```
 const clearFilter = (refreshUI: boolean = true) => {
    window.sessionStorage.removeItem('filter');
    if (refreshUI) {
      props.refetch();
      setCurrentFilter(emptyFilter);
    }
  };
```
But I do not see this in the current Changes. I am not confident that this would work and in that case I am good with landing the current solution

abishekvashok:
Hey @r0rshark I had to modify his suggestion a bit to make it work, I can see this in the diff:
```
export function clearFilter(
  refetch: any = false,
  setCurrentFilter: any = false,
  currentFilter: FilterDescription = emptyFilter,
): void {
  if(!refetch) {
    window.sessionStorage.removeItem('filter');
  } else  {
    setCurrentFilter(emptyFilter);
    // workaround for https://github.com/apollographql/react-apollo/issues/3709
    // implemented in clearAndRefetch function makes us implement this hacky
    // solution. Otherwise, the app just keeps loading and does nothing.
    refetch(filterToVariables(emptyFilter));
    refetch(filterToVariables(currentFilter));
  }
}
```

0xedward:
@abishekvashok The intention was to merge the two functions in this PR and PR#29, while cleaning up the function signature and reducing the number of refetch calls. 

The preliminary testing I did when I suggested [the code change](https://github.com/facebook/sapp/pull/28#discussion_r675192680) seemed to indicate it would work for clearing the current filter and refreshing the UI. What wasn't working that required modification to the code suggestion? I'm happy to help with fixing it :)

abishekvashok:
Upss I am sorry for not explaining it!

> What wasn't working that required modification to the code suggestion? 

Since the function was to be called also from  run.js in the other pr, we needed to move it out of the component and export it as a single function - hence, `export function` instead of `const` (Export const functionname works too).

Then, since it is outside the component, the function couldn't access the 3 variables (props.refetch, currentFilter and setCurrentFilter), so passing them as (optional) arguments (this is done in a 'hacky way', so like when refetch is passed, it is a function hence the if statement evaluates to false and the else block is executed, if there is no such argument as in the other PR, the !refetch becomes true and it executes the if block). Also since it is an argument and not a prop, props.refetch becomes refetch, etc.

Regarding the order of refetch and setCurrentFilter, I guess it wouldn't work if we call refetch and set the current filter as empty filter? I can try that and see if it works - if so we would only need one refetch call. Else, we would need two.

abishekvashok:
> Regarding the order of refetch and setCurrentFilter, I guess it wouldn't work if we call refetch and set the current filter as empty filter? I can try that and see if it works - if so we would only need one refetch call. Else, we would need two as is in this pr.

I tried this and my conclusion is that we will need two refetch calls as `sometimes` this doesn't work. As a hint, try using a filter that passes everything through (say you have only 5 issues with code 1000 in a run, and you apply a filter with code 1000) and try clearing it, you will get this loading animation:
Screenshot 2021-07-29 at 2 56 24 PM

0xedward:
>Since it is outside the component, the function couldn't access the 3 variables (props.refetch, currentFilter and setCurrentFilter), so passing them as (optional) arguments

Ah that makes sense, I notice we [export the `Filter` object in `Filter.js`](https://github.com/facebook/sapp/blob/master/sapp/ui/frontend/src/Filter.js#L1051). I'm guessing it's not possible to access the `clearFilter` property from another file?

>Regarding clearing the session storage, it is only needed when we switch from run-to-run because the issues page actually initialises it as an empty filter if there is no such value in storage while the component loads.

> I tried this and my conclusion is that we will need two refetch calls as sometimes this doesn't work. As a hint, try using a filter that passes everything through (say you have only 5 issues with code 1000 in a run, and you apply a filter with code 1000) and try clearing it, you will get this loading animation that goes on forever,

The intention behind suggesting to call `window.sessionStorage.clear()` was to try reduce the number of `refetch` calls we had to make, since removing the `filter` item from `sessionStorage` and then readding `filter` with an empty filter theoretically should be equivalent to changing the `filter` item in `sessionStorage` to an empty filter. But it looks like with your testing, it doesn't always work (e.g. when the entire list of issues needs to be refreshed).

Would you have any thoughts or recommendations here, @dkgi? If not, I think we are cool with landed this as is. 

Thanks for working on this and doing the additional testing! :)

abishekvashok:
> Ah that makes sense, I notice we export the Filter object in Filter.js. I'm guessing it's not possible to access the clearFilter property from another file?

We can export one component using `export default` so it can be imported in the other files as:
`import component-name from 'filename'`

Here, Filter is exported. But, react provides us another way to export more than one object via the export keyword.
So by specifying `export function ...` we are exporting the clear filter function and it can be imported as:
`import {functionName} from 'filename'`

:)

abishekvashok:
Fixed merge conflicts, and updated the position to reflect the new filter UI:
Screenshot 2021-08-03 at 10 04 52 AM

PR #30, opened 2021-07-21
Adds support for Issue Status

abishekvashok:
Adds support for displaying, editing, and filtering with issue status.

Adds mutation to update issue status relating to a particular issues,
resolution functions, modifies definitions to incoporate issue status,
and adds filtering support to filter by particular issue status.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/35

### DIFF
--- a/sapp/ui/frontend/src/Issue.js
+++ b/sapp/ui/frontend/src/Issue.js
@@ -9,11 +9,14 @@
  */
 
 import React, {useState} from 'react';
+import {useMutation, gql} from '@apollo/client';
 import {
   Card,
   Col,
+  Modal,
   Popover,
   Row,
+  Select,
   Skeleton,
   Tag,
   Tooltip,
@@ -30,6 +33,7 @@ import Source from './Source.js';
 import {Documentation} from './Documentation.js';
 import {HumanReadable} from './HumanReadable';
 
+const {Option} = Select;
 const {Text, Link} = Typography;
 
 function ShowMore(
@@ -130,10 +134,63 @@ export type IssueDescription = {
   source_names: $ReadOnlyArray<string>,
   sinks: $ReadOnlyArray<string>,
   sink_names: $ReadOnlyArray<string>,
+  status: string,
   features: $ReadOnlyArray<string>,
   min_trace_length_to_sources: number,
   min_trace_length_to_sinks: number,
   is_new_issue: boolean,
+};
+
+export const statusMap = {
+  "uncategorized": "Uncategorized",
+  "bad_practice": "Bad practice",
+  "false_positive": "False positive",
+  "valid_bug": "Valid bug",
+  "do_not_care": "Do not care",
+};
+
+const StatusSelect = (
+  props: $ReadOnly<{
+    issue_id: number,
+    status: string,
+  }>,
+): React$Node => {
+  let allStatuses = [];
+  for(let key in statusMap) {
+    allStatuses.push(<Option value={key}>{statusMap[key]}</Option>);
+  };
+
+  const updateStatusMutation = gql`
+    mutation UpdateStatus($id: ID!, $status: String!) {
+      update_status(input: {id: $id, status: $status}) {
+        clientMutationId
+      }
+    }
+  `;
+
+  const [updateStatus, {error: statusUpdateError}] = useMutation(
+    updateStatusMutation
+  );
+
+  const updateStatusFunc = (value) => {
+    updateStatus({variables: {id: props.issue_id, status: value}})
+  };
+
+  if(statusUpdateError) {
+    Modal.error({
+      title: 'Unable to update status',
+      content: statusUpdateError.toString()
+    });
+  }
+
+  return(
+    <Select
+      size='small'
+      defaultValue={props.status}
+      onChange={updateStatusFunc}>
+      {allStatuses}
+    </Select>
+  );
 };
 
 export function Issue(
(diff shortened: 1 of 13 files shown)


abishekvashok:
Changes:

https://user-images.githubusercontent.com/8947010/126478871-3e432689-b58f-4075-9ab9-3044b50a60f5.mov

gbleaney (changes requested):
Just to make sure I understand correctly, the database schema already supported statuses, and this is just piping it up to the UI and back, right?

0xedward:
> Just to make sure I understand correctly, the database schema already supported statuses, and this is just piping it up to the UI and back, right?

@gbleaney - the filter object is stored in the `filters` table like this:
| name      | description | json |
| ----------- | ----------- | --------- |
| Command and Code Injection | High signal filter for 5001 Shell Injection | {"features": [{"mode": "none of", "features": ["always-type:scalar", "always-type:bool", "via:shell_escape", "always-via:shell_escape"]}], "codes": [5001], "traceLengthFromSources": [0, 3], "traceLengthToSinks": [0, 4]} |

So [any attribute that's not `name` or `description`](https://github.com/facebook/sapp/blob/master/sapp/filter.py#L64-L65) will be serialize and stored in the `json` column

For `IssueStatus`, it seems like we [already have a column in models.py](https://github.com/facebook/sapp/blob/HEAD/sapp/models.py#L586-L592)

sapp/ui/filter_predicates.py
```diff
@@ -85,7 +85,8 @@ def __init__(self, column: Union[Column[_T], DBID], items: Sequence[_T]) -> None
 
     def apply(self, query: Query[_Q]) -> Query[_Q]:
         # pyre-ignore: SQLAlchemy too dynamic.
-        return query.filter(or_(*[self._column.like(item) for item in self._items]))
```
0xedward, line 88: What's the reason for this deletion? It seems to be exactly the same as the code block two lines below it:
```
x = or_(*[self._column.like(item) for item in self._items])
return query.filter(x)
```

dkgi:
Ultra-nit: could we change the vertical alignment of the dropdown and/or reduce the height of the dropdown. Right now it looks a bit unbalanced in the form.

0xedward (changes requested):
It looks great! Just had one question

sapp/ui/filter_predicates.py
```diff
@@ -144,6 +145,8 @@ def apply(self, issues: List[IssueQueryResult]) -> List[IssueQueryResult]:
                 for issue in issues
                 if len(issue.sink_kinds & self._parameter_list) > 0
             ]
+        elif self._parameter_type == "statuses":
+            return [issue for issue in issues]
```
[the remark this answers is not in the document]
abishekvashok: Not meant to be there. I messed this file up! Sorry!

sapp/ui/filter_predicates.py
abishekvashok: None, sorry this file wasn't to be added to the commit

sapp/ui/frontend/src/Issue.js
```diff
@@ -130,12 +134,61 @@ export type IssueDescription = {
...
   is_new_issue: boolean,
 };
 
+const StatusSelect = (
+  props: $ReadOnly<{
+    issue_id: number,
+    status: string,
+  }>,
+): React$Node => {
+  const allStatuses = [
```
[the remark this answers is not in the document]
abishekvashok: Done

abishekvashok:
Hey, thank you for the reviews guys
@0xedward @gbleaney @dkgi removed that unwanted diff, which I used for testing and added user friendly status es by capitalising the first letter and replacing _ with whitespace in both the issues and filters.

Also, to the suggestion to make it smaller: I've made it smaller!

Here's a small video that reflects the state of this diff:

https://user-images.githubusercontent.com/8947010/126537266-9a464de1-b932-4f50-84f5-661a2a17603c.mov


Thank you guys again for taking your time to review and give out suggestions, really appreciate it :)

r0rshark:
> Just to make sure I understand correctly, the database schema already supported statuses, and this is just piping it up to the UI and back, right?

Correct. Before creating the task I verified that the field was already defined and we had the data in the db as well (status is uncategorized by default)
Screenshot 2021-07-22 at 10 41 01

sapp/ui/frontend/src/Issue.js
```diff
@@ -130,12 +134,62 @@ export type IssueDescription = {
...
+    "Uncategorised",
```
r0rshark, line 151: @abishekvashok What do you think about having a dictionary which maps the status in the same version as the backend (IssueStatus.BAD_PRACTICE bad_practice) with a human readable version. Something like 
{
"bad_practice": "Bad Practice",
...
}
In this way we can show to the user the human readable version but send to the graphQL query the same representation defined in the backend (IssueStatus )
This would avoid having to do all the `status.replace(" ", "_").lower()` conversions between the human readable and the backed IssueStatus representation in the python code which I think can be very error prone

r0rshark (changes requested):
Looks good just a couple of suggestions :)

sapp/ui/tests/interactive_test.py
```diff
@@ -1786,6 +1787,7 @@ def testCreateIssueOutputStringNoSourcesNoSinks(self):
             source_kinds=None,
             sink_names=None,
             sink_kinds=["sink1", "sink2"],
+            status=IssueStatus.UNCATEGORIZED,
```
r0rshark, line 1790: Can we add a test similarly to what we did for [testListIssuesFilterCodes]( https://github.com/facebook/sapp/blob/e30857392d6ab2f8e399274d41b7a957678ea511/sapp/ui/tests/interactive_test.py#L205)

sapp/ui/frontend/src/Issue.js
abishekvashok: Yep makes sense!

sapp/ui/tests/interactive_test.py
abishekvashok: I will write one and get back :)

sapp/ui/frontend/src/Issue.js
0xedward, line 151:
```suggestion
    "Uncategorized",
```

0xedward (changes requested):
Small suggestion to match human readable strings with [`IssueStatus` values](https://github.com/facebook/sapp/blob/master/sapp/models.py#L531)

abishekvashok:
@r0rshark @0xedward the requested changes have been made :)

sapp/ui/tests/interactive_test.py
abishekvashok: Done :)

abishekvashok:
Done :)

PR #31, opened 2021-07-28
Adds ability to export existing filters

abishekvashok:
Adds ability to export an existing filter from the UI. Adds export
button to the filter options that when clicked downloads the filter json
file thereby exporting it.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/39

### DIFF
--- a/sapp/ui/filters.py
+++ b/sapp/ui/filters.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, List, Tuple, Optional
 
 import graphene
 import sqlalchemy
+from flask.views import View
 from sqlalchemy.orm import Session
 
 from .. import models
@@ -164,6 +165,31 @@ def export_filter(
                 LOG.info(f"`{filter_name}` has been exported to `{output_filter_path}`")
             else:
                 print(storedfilter_instance.to_file())
+        except sqlalchemy.exc.OperationalError:
+            LOG.error(
+                "Error: Database disk image is malformed or "
+                "database has not been initialized properly."
+                "Please recreate your SAPP db."
+            )
+            raise
+
+
+class ServeExportFilter(View):
+    def __init__(self, session: Session) -> None:
+        self.session = session
+
+    # pyre-fixme[14]: Inconsistent override
+    def dispatch_request(self, filter_name: str) -> str:
+        try:
+            record = (
+                self.session.query(FilterRecord)
+                .filter(FilterRecord.name == filter_name)
+                .one_or_none()
+            )
+            if not record:
+                raise FilterNotFound(f"`{filter_name}` does not exist")
+            storedfilter_instance = StoredFilter.from_record(record)
+            return storedfilter_instance.to_file()
         except sqlalchemy.exc.OperationalError:
             LOG.error(
                 "Error: Database disk image is malformed or "
--- a/sapp/ui/frontend/src/Filter.js
+++ b/sapp/ui/frontend/src/Filter.js
@@ -31,6 +31,7 @@ import {
   Typography,
 } from 'antd';
 import {
+  ExportOutlined,
   FieldTimeOutlined,
   FilterOutlined,
   PlusOutlined,
@@ -1070,6 +1071,19 @@ export const FilterControls = (props: {
             onClick={() => setDeleteModalVisible(true)}
             icon={<DeleteOutlined />}>
             Delete
+          </Menu.Item>
+          <Menu.Item
+            disabled={currentFilter?.name === undefined}
+            icon={<ExportOutlined />}>
+            {(currentFilter?.name === undefined) ? "Export" :
+              <a
+                href={`/export_filter/${currentFilter.name}`}
+                download={`${currentFilter.name}.json`}
+                target='_blank'
+                rel='noopener noreferrer'>
+                Export
+              </a>
+            }
           </Menu.Item>
         </SubMenu>
         <SubMenu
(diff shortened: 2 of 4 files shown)


abishekvashok:
Changes in action:


https://user-images.githubusercontent.com/8947010/127311603-c59b1266-721a-42b4-9bb9-5e5f206feaf5.mov


https://user-images.githubusercontent.com/8947010/127311606-abb6a0cd-2535-438e-998d-07eebea8f6fb.mov


(Had to split the video into two because of 100mb file size limitation GitHub imposes)

r0rshark:
Nice can we also add another PR for supporting this from the UI with the ability to export one filter or all of them together?

abishekvashok:
> Nice can we also add another PR for supporting this from the UI with the ability to export one filter or all of them together?

@r0rshark sure, we are lacking importing a json via UI as well, would you guys be keen ofr both of them in a single PR?

r0rshark:
@abishekvashok sorry I poorly explained the issue in the description, we have an internal diff for supporting the export of the filters from the command line but we do not have AFAIK (@0xedward correct me if I am wrong) this function supported in the SAPP UI. Can we change this diff on implementing the importing/exporting from the SAPP UI? Thanks :)

abishekvashok:
Caution for importing - we would need to process files, so an additional flask endpoint has to be added - I can proceed if it's fine, for now, I will change this to export via UI :)

abishekvashok:
@r0rshark done, changed the branch history, now it's export in the UI:

https://user-images.githubusercontent.com/8947010/127487883-a6aa94e4-8e7d-4003-a4a1-58241f4893c5.mov

abishekvashok:
@r0rshark I have found a way to process files within React itself, made a pr for that: https://github.com/facebook/sapp/pull/34 :)

r0rshark:
> It's okay @r0rshark
> Caution for importing - we would need to process files, so an additional flask endpoint has to be added - I can proceed if it's fine,
> 
> For now, I will change this to export via UI :)

Great that's exactly what I had in mind!

abishekvashok:
Fixed merge conflicts, here's how it looks now with the new UI changes:
Screenshot 2021-08-03 at 10 28 57 AM

r0rshark:
@abishekvashok  in https://github.com/facebook/sapp/commit/66895fa864b6a5a04f6bed534293f165570f8bcb @0xedward  implemented a server side function which we use to export the filter from the sapp cli. In order to keep the same logic both in the UI and in the cli I think it's worth refactoring the code to make the Frontend get the filter content from the StoredFilter.to_file() backend function. What do you think?

abishekvashok:
> @abishekvashok in 66895fa @0xedward implemented a server side function which we use to export the filter from the sapp cli. In order to keep the same logic both in the UI and in the cli I think it's worth refactoring the code to make the Frontend get the filter content from the StoredFilter.to_file() backend function. What do you think?

Makes sense :)

abishekvashok:
@r0rshark done!

https://user-images.githubusercontent.com/8947010/129381043-2add26b0-f084-4374-bc83-f395af28256d.mov

Reused much of @0xedward 's code, thanks for that!

sapp/ui/filters.py
```diff
@@ -171,6 +172,32 @@ def export_filter(
             raise
 
 
+class ServeExportFilter(View):
```
gbleaney, line 175: The rest of the app runs over graphql. Why did we need to switch to a view here? It's not a blocker to landing this PR, per se, but I want to be sure we're not adding a new paradigm without a good reason. Each new way of querying the server just increases the maintenance overhead a bit.

gbleaney (changes requested):
TY for the PR as usual! You've been a huge help this summer. Only the unreachable return comment is actually a must fix.

sapp/ui/filters.py
```diff
@@ -171,6 +172,32 @@ def export_filter(
...
+                return f"`{filter_name}` does not exist`"
+            storedfilter_instance = StoredFilter.from_record(record)
+            return storedfilter_instance.to_file()
+        except sqlalchemy.exc.OperationalError:
+            LOG.error(
+                "Error: Database disk image is malformed or "
+                "database has not been initialized properly."
+                "Please recreate your SAPP db."
+            )
+            raise
```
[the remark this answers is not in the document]
abishekvashok: 😁

sapp/ui/filters.py
```diff
@@ -171,6 +172,32 @@ def export_filter(
...
+    def dispatch_request(self, filter_name: str) -> str:
+        try:
+            record = (
+                self.session.query(FilterRecord)
+                .filter(FilterRecord.name == filter_name)
+                .one_or_none()
+            )
+            if not record:
+                raise FilterNotFound(f"`{filter_name}` does not exist")
+                return f"`{filter_name}` does not exist`"
```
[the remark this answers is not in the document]
abishekvashok: You are right! Removed it!

sapp/ui/filters.py
abishekvashok: Okay, so sorry for finding a new approach. We can do it this PR three ways:
 - use the filter on the frontend (this seems the best approach as we already query the saved filter before export) (this was migrated to use this approach)
 - use filter along with backend as shown in this PR
 
Making an additional backend query on click of `a` to download file doesn't seem to work as often the file gets downloaded before the results arrive from the backend :/ But however, there is an option to display an additional pop-up saying the filter is ready, but then that kinda makes the UX bad. I was out of my Witts here and went with the additional view which imo is also kinda bad as you mentioned.

r0rshark: I am fine with this approach. The view is more suited to deliver files since it easily give us a URL we can embed in a <a> tag to download the file (GET /export_filter/whatever). As @abishekvashok  points out using GraphQL the solution would be more involved requiring onclick event submitting a POST request to graphQL and then locally downloading the file. Compared to the solution 1. this will guarantee us that the export functionality is common between the cli and the UI and doesn't rely on the clientside implementation of the currentFilter React JS props :)

PR #32, opened 2021-07-28
Mirror the Facebook internally used linter/formatter behaviour in the open source repository.

abishekvashok:
Mirror the Facebook internally used linter/formatter behaviour in the
open source repository.

Pins to black and usort versions used in the internal linter and runs
black, flake8, and usort in seperate runner instances.

Adds pyproject.toml with identical usort configuration as used
internally.

Modifies flake8 configuration to match the one used internally.

Fix errors reported by flake8, black, and usort.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/40

### DIFF
--- a/sapp/pipeline/mariana_trench_parser.py
+++ b/sapp/pipeline/mariana_trench_parser.py
@@ -24,12 +24,6 @@ from .. import pipeline as sapp
 from ..analysis_output import AnalysisOutput, Metadata
 from .base_parser import BaseParser
 
-try:
-    # type: ignore
-    from ...facebook.lib import configuration
-except Exception:
-    from .. import configuration
-
 if sys.version_info >= (3, 8):
     from typing import Literal
 else:
--- a/sapp/ui/frontend/package.json
+++ b/sapp/ui/frontend/package.json
@@ -76,10 +76,16 @@
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",
     "test": "node scripts/test.js",
-    "flow": "flow"
+    "flow": "flow",
+    "lint": "npx eslint . --max-warnings=0"
   },
   "eslintConfig": {
-    "extends": "react-app"
+    "extends": "react-app",
+    "rules": {
+      "eqeqeq": "off",
+      "strict": "off",
+      "default-case": "off"
+    }
   },
   "browserslist": {
     "production": [
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,20 @@
+[flake8]
+select = B,C,E,F,P,T4,W,B9
+max-line-length = 80
+# E127, E128 are hard to silence in certain nested formatting situations.
+# E265, E266 talk about comment formatting which is too opinionated.
+# E402 warns on imports coming after statements. There are important use cases
+# like demandimport (https://fburl.com/demandimport) that require statements
+# before imports.
+# E501 is not flexible enough, we're using B950 instead.
+# E722 is a duplicate of B001.
+# P207 is a duplicate of B003.
+# W503 talks about operator formatting which is too opinionated.
+# E203 conflicts with Black
+ignore = E127, E128, E265, E266, E402, E501, E722, P207, W503, E203, T484
+max-complexity = 50
+exclude =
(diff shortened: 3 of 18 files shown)


r0rshark (commented):
Discussed about adding ESlint support as well but I am OK doing it in a separate PR :)

abishekvashok:
@r0rshark working at it atm, I don't want to make a separate commit for it.

r0rshark:
> @r0rshark working at it atm, I don't want to make a separate commit for it.

Great! that's better

abishekvashok:
@r0rshark done :)

sapp/ui/frontend/package.json
```diff
@@ -79,7 +79,12 @@
     "flow": "flow"
   },
   "eslintConfig": {
-    "extends": "react-app"
+    "extends": "react-app",
+    "rules": {
+      "eqeqeq": "off",
```
r0rshark, line 84: Why we disabled these?

r0rshark (commented):
Just one clarification

0xedward (changes requested):
It's a great idea to include ESLint here! In order to keep the results from manually running ESLint consistent with the GitHub actions linter run, could we add an npm command (e.g. `npm run lint`) do run ESLint with the flags we prefer?

i.e. Adding an entry to `"scripts"` in `package.json`

sapp/ui/frontend/package.json
abishekvashok: Javascript has two operators ==, === one compares values and the other compares value and type. Both are useful. JSLint just prefers the use of === over == and complains when we use ==. Since changing all == to === can break things, I thought it is nice to avoid the check.

https://stackoverflow.com/questions/3735939/jslint-expected-and-instead-saw

But yeah, if we want to just use === throughout, I am okay to remove it :)

abishekvashok:
@0xedward done :)

r0rshark:
@abishekvashok  it looks like there is a bug https://github.com/facebook/sapp/pull/32/checks?check_run_id=3227539011 since in the CI we are picking up the wrong path `npm ERR! enoent ENOENT: no such file or directory, open '/home/runner/work/sapp/sapp/package.json'` instead of `'/home/runner/work/sapp/sapp/ui/frontend/package.json'`
`

abishekvashok:
The actions don't preserve current working directory between steps. This should be okay now.

sapp/pipeline/mariana_trench_parser.py
```diff
@@ -24,12 +24,6 @@
 from ..analysis_output import AnalysisOutput, Metadata
 from .base_parser import BaseParser
 
-try:
-    # type: ignore
-    from ...facebook.lib import configuration
-except Exception:
-    from .. import configuration
-
```
0xedward, line 32: Interesting that `flake8` seems to think these imports are unused, since [`from .. import configuration` is a valid path](https://github.com/facebook/sapp/blob/master/sapp/configuration.py). Maybe because the imports are wrapped in a try-except block?

Anyway, we need to have the linter ignore these lines, because if we delete these imports, we will break `sapp analyze` pipeline for internally and externally for one of our static analysis tools

abishekvashok: It was removed because it was unused. I will add it back if it's needed :)

abishekvashok:
@0xedward done :)

sapp/pipeline/mariana_trench_parser.py
0xedward: Oh, my bad! I'm not sure how I missed this earlier, but it doesn't look like the file is used at all anymore

So we can leave the deletion of the entire try-except block here and can also delete [sapp/configuration.py](https://github.com/facebook/sapp/blob/master/sapp/configuration.py) as well. Sorry about missing this on my first pass :(

abishekvashok: @0xedward it's fine, I will just revert and check and delete the file as well :)

abishekvashok:
@0xedward done :)

PR #33, opened 2021-07-29
Allow user to update a saved filter without clear

abishekvashok:
Currently, to update a saved filter, the user has to clear existing
filters, enter in the filter parameters, click save, and then enter the
existing filter name to update it.

Changed the workflow such that the save button isn't disabled when an
existing filter is selected and convey the filtername to the save model.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/32

### DIFF
--- a/sapp/ui/frontend/src/Filter.js
+++ b/sapp/ui/frontend/src/Filter.js
@@ -754,6 +754,12 @@ const SaveFilterModal = (
 ): React$Node => {
   const [form] = Form.useForm();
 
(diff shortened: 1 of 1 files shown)


abishekvashok:
Before this change:
Screenshot 2021-07-29 at 6 12 43 PM

After the change:

https://user-images.githubusercontent.com/8947010/127494034-87b288c7-8ca2-4981-8992-26745876c132.mov

PR #34, opened 2021-07-29
Adds ability to import Filter(s) via UI

abishekvashok:
Adds feature to import one or more filter json files to the database via
the react UI.

Adds a draggable component through which filter json files can be
dragged into or selected with the help of the on system file picker.
Only json files are alllowed and the other sort of files are skipped.
File reading is done on the client side, in the react UI, the contents
are then passed to the graphQL database via the existing save filter
mutation.

Modifies the mutation associated function so that they won't load the
last imported filter but retain such functionality on saving a new
filter as before.


abishekvashok:
PR inspired from this comment: https://github.com/facebook/sapp/pull/31#issuecomment-888998259
Thought, it's wise to keep import and export separated.

Changes in action:

https://user-images.githubusercontent.com/8947010/127543087-1869f594-3e5d-41a3-868a-6503361950ac.mov

sapp/ui/frontend/src/Filter.js
```diff
@@ -780,7 +784,57 @@ const SaveFilterModal = (
...
+    filterList,
+  };
+
+  const onImport = () => {
+    let count = filterList.length;
+    filterList.forEach(filterFile => {
+      let reader = new FileReader();
+      reader.onload = (e) => {
+        let filter: FilterDescription = JSON.parse(e.target.result);
+        if(filter["name"] !== undefined &&
```
dkgi, line 812: Formatting here looks off (space after `if`), this might be something we'd want to add to the open source workflow (I'm not sure what tool we're using internally, @0xedward maybe?).

sapp/ui/frontend/src/Filter.js
```diff
@@ -780,7 +784,57 @@ const SaveFilterModal = (
...
+          filter["description"] !== undefined &&
+          Object.keys(filter).length > 2
+        ) {
```
dkgi, line 815: nit: prefer early aborts, i.e. something along the lines of

```
if (!(<what we want>)) {
  // error handling, are there error dialog helpers in the React library we're using?
}
// rest of the code
```

dkgi (changes requested):
Oh this looks very nice 👍 

Left couple of inline comments.

For the UX: one of the things that I found slightly confusing in the video is that there is no noticeable change in the UI after the filters get imported. Not sure how to best address this. One way would be to not support batch operations and automatically select the imported filter. That would be somewhat limiting but probably less confusing. Curious what others think though.

abishekvashok:
@dkgi thanks for looking into this :) I will work on the suggestions,

> For the UX: one of the things that I found slightly confusing in the video is that there is no noticeable change in the UI after the filters get imported.

Meanwhile, what do you think of an antd success [message](https://ant.design/components/message/)

abishekvashok:
@dkgi fixed two suggestions. Also added a message success to display the imported filter count :)
Screenshot 2021-08-02 at 4 13 30 PM

sapp/ui/frontend/src/Filter.js
abishekvashok: Yep we use message. Reusing it :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -780,7 +784,65 @@ const SaveFilterModal = (
...
+          saveFilter({
+            variables: {
+              name: name,
+              description: description,
+              json: JSON.stringify(filter),
+            },
+          });
+        }
+        count--;
+        if(count === 0) {
```
0xedward, line 833: Do we need to decrement and keep track of `count` here? Would we be able to have the following code block after the for each loop ends instead?
```javascript
          props.hideImportModal();
          let imported = filterList.length - skipped;
          if(imported !== 0) {
            message.success(`Imported ${imported} filters`);
          }
```

abishekvashok: Yes makes better sense! I will try this out.

sapp/ui/frontend/src/Filter.js
```diff
@@ -780,7 +784,65 @@ const SaveFilterModal = (
...
+  };
+
+  const onImport = () => {
+    let count = filterList.length;
+    let skipped = 0;
+    filterList.forEach(filterFile => {
+      let reader = new FileReader();
+      reader.onload = (e) => {
+        let filter: FilterDescription = JSON.parse(e.target.result);
+        if (filter["name"] === undefined ||
```
0xedward, line 813: Instead of filtering for which files in `filterList` we can import in the same loop as importing filters, could we use `Array.prototype.filter()` to filter invalid files first, then run the for each loop below on the valid list of filters?

The callback function we use in `Array.prototype.filter()` to validate if a filter file is valid can be reused to perform validation before we export filters in https://github.com/facebook/sapp/pull/31

sapp/ui/frontend/src/Filter.js
0xedward: I'm not sure if our internal linter catches whitespace issues, but there's a [PR](https://github.com/facebook/sapp/pull/32) to add ESLint where we can see if we can include a rule to cover this though.

The ESLint config in the PR extends the [recommended rule set](https://eslint.org/docs/rules/) so I think it should cover this case for us

sapp/ui/frontend/src/Filter.js
abishekvashok: Hey nice idea but the problem is the fileReader which is kinda not straight flowing with its onload function. So I did try using it on the beforeUpload function, but even with Promises I couldn't get it to work properly, hence I put the validation here.

sapp/ui/frontend/src/Filter.js
abishekvashok: Done

sapp/ui/frontend/src/Filter.js
0xedward: > Nice idea but the problem is the fileReader which is kinda not straight flowing with its onload function. 

@abishekvashok Sorry I'm not following - what was the issue with `FileReader`'s `onload` property? Was the issue related to the `load` event not firing when you included it in the `callback` function for `Array.prototype.filter()`? 

Would something roughly like this not work?
```javascript
const filterInvalidFilterFiles(filterFile) {
	let reader = new FileReader();
	reader.onload = (e) => {
        let filter: FilterDescription = JSON.parse(e.target.result);
        if (filter["name"] !== undefined &&
        	      filter["description"] !== undefined &&
          Object.keys(filter).length > 2) {
        	return true;
	}
    	message.error(`Invalid filter file, skipping import of ${filterFile.name}`);
    	return false;
	}
}

const onImport = () => {
       ...
       filterList.filter(filterInvalidFilterFiles)

```

0xedward (changes requested):

sapp/ui/frontend/src/Filter.js
abishekvashok: Yep, this will return true everytime, I pushed this in beforeUpload but I was always getting true.

abishekvashok: Hmm return inside onload, makes sense, I will try this :) Thank you

abishekvashok: I tried it inside of beforeUpload, and appending the result to filterList to so that the reader can be used only once, but it doesn't work - the same result :/

abishekvashok: I will try a separate filter function now, as provided, but we will need to read the file two times!

abishekvashok: @0xedward Did try, not working always returning true :/

abishekvashok: Uff, I was being dump :| the problem is that reader.onload is a function. So when we return true inside of the function -> it is the value returned from reader.onload() and not our `removeInvalidFilterFiles(filterFile)` function. So whatever reader.onload() does return, our `removeInvalidFilterFiles(filterFile)` function always returns void.

abishekvashok:
Hey @0xedward , for some reason GitHub pushes my comment backwards even though it was made after your review. I though it would be nice to restate in case you missed it :)

> @0xedward Uff, I was being dump :| sorry. The problem is that reader.onload is a function. So when we return true inside of the function -> the value is returned from reader.onload() and not our removeInvalidFilterFiles(filterFile) function. So be it whatever reader.onload() does return, our removeInvalidFilterFiles(filterFile) function always returns void here. I tried checking if there's a work around but it seems reader.onload() 's value isn't stored anywhere. Same thing occurred with promises.
https://developer.mozilla.org/en-US/docs/Web/API/FileReader

0xedward:
> Hey @0xedward , for some reason GitHub pushes my comment backwards even though it was made after your review. I though it would be nice to restate in case you missed it :)

@abishekvashok sorry, I totally missed this! Thanks for giving the suggestion a try :) The idea here was to separate filtering out invalid files from importing filters. 

Here are some more alternatives:
1. We could make another function that creates a new FileReader instance and returns a copy of the list of valid filters. We can call that function in the beginning of `onImport`.
2. Since `saveFilter` does validation, we can remove the code block that checks if a filter is valid and wrap the `saveFilter` call in a try-catch

Not sure if others think those alternatives would be worth doing here, so I'll import the PR and see what others think

abishekvashok:
@0xedward thanks for the clarification. After some time of researching on with antd, I think I found a way to isolate the validation and uploading! We required a custom dummy request with a zero tick timeout as for jest, done it! 🥳

sapp/ui/frontend/src/Filter.js
```diff
@@ -746,12 +751,15 @@ const FilterForm = (props: {
...
 
-const SaveFilterModal = (
+const SaveAndImportFilterModals = (
   props: $ReadOnly<{|
     currentFilter: FilterDescription,
-    visible: boolean,
-    hide: void => void,
-    onSave: FilterDescription => void,
+    setCurrentFilter: FilterDescription => void,
+    saveModalVisible: boolean,
```
dkgi, line 758: nit: I think by convention we use `is` as prefix for boolean flags (disclaimer: I'm highly unsure about whether that is actually a thing)

sapp/ui/frontend/src/Filter.js
```diff
@@ -788,7 +793,77 @@ const SaveFilterModal = (
...
+              filterList = [...filterList, filter];
+              filterNamesList = [...filterNamesList, `${filter.name}.json`]
+              resolve(filter);
+            }
+          };
+          reader.readAsText(filterFile);
+        }
+      });
+    },
+    customRequest: options => {
```
dkgi, line 836:
```suggestion
    customRequest: {onSuccess, onError, file} => {
```

dkgi (commented):
As far as I can tell this looks reasonable :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -788,7 +793,77 @@ const SaveFilterModal = (
...
+  let filterNamesList = [];
+
+  const isValidFilter = (filter: FilterDescription): Boolean => {
+    if (filter.name === undefined ||
+        filter.description === undefined ||
+        Object.keys(filter).length < 3
+    ) {
+      return false;
+    }
+    return true;
```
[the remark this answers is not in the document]
abishekvashok: Hell yeah!

sapp/ui/frontend/src/Filter.js
abishekvashok: I tried searching for `isBoolean` and `isboolean` in the repo, but couldn't find it :/ are you sure that it was what you meant 🙃

Also, we could defy convention and make this `isBoolean` if you really want, or we could do away with the type declaration all together!

sapp/ui/frontend/src/Filter.js
abishekvashok: Never knew this! Thanks for teaching ;)

abishekvashok: Are you sure about this? Maybe, I am missing something. But this doesn't work 🙃 Can you provide more information pleaseee?

abishekvashok:
@dkgi unsure about your last two comments. Thanks for your time. Made changes to the function :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -788,7 +793,73 @@ const SaveFilterModal = (
...
+        if (filterFile.type !== 'application/json') {
+          message.error(`${filterFile.name} is not a json file`);
+          resolve(Upload.LIST_IGNORE);
+        } else {
+          let reader: FileReader = new FileReader();
+          reader.onload = (e) => {
+            let filter: FilterDescription = JSON.parse(e.target.result);
+            if (!isValidFilter(filter)) {
+              message.error(`${filterFile.name} is an invalid filter file`);
+              resolve(Upload.LIST_IGNORE);
```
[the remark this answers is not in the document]
abishekvashok: Yes, actually the result of the promise is never looked after. I returned `Upload.LIST_IGNORE` as I thought that would be more readable - but will switch to just resolve. Also this is why the custom request I used to call onError and the filterNamesList is used!

sapp/ui/frontend/src/Filter.js
```diff
@@ -788,7 +793,73 @@ const SaveFilterModal = (
...
+            } else {
+              filterList = [...filterList, filter];
```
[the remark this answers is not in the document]
abishekvashok: Sure :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -788,7 +793,73 @@ const SaveFilterModal = (
...
+              filterNamesList = [...filterNamesList, `${filter.name}.json`]
+              resolve(filter);
+            }
+          };
+          reader.readAsText(filterFile);
+        }
+      });
+    },
+    customRequest: options => {
```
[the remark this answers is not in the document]
abishekvashok: Yeah! Thanks for this - literally couldn't get it via googling - but this helps :)

abishekvashok:
@gbleaney thank you for your time. Made the requested changes :)

PR #35, opened 2021-08-02
Simplify development setup for sapp

abishekvashok:
Currently, we need to rebuild the react app each time when we make any
changes to the UI and hot-reload is unavialable. Changes that by
introducing hot-reload by running the app on 3000 port and the flask
server on port 5000 so that hot-reload is available for both the
frontend and backend.

Adds CORS headers for the same to enable communication between both
frontend and backend using flask-cors library while on debug
(development) mode. Also adds flask-cors to requirements.txt

Updates the docs to specify the same.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/41

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,9 +259,9 @@ $ source ~/.venvs/sapp/bin/activate
 (sapp) $ pip3 install -r requirements.txt
 ```
 
-Run the flask server:
+Run the flask server in debug mode:
 ```shell
-(sapp) $ python3 -m sapp.cli server
+(sapp) $ python3 -m sapp.cli server --debug
 ```
 
 Parse static analysis output and save to disk:
@@ -269,17 +269,21 @@ Parse static analysis output and save to disk:
 (sapp) $ python3 -m sapp.cli analyze taint-output.json
 ```
 
-If you make any changes to files under `sapp/ui/frontend/*`, you will need to run `npm install` once to install dependencies and `npm run-script build` each time you make changes before running the flask server to see the changes you made reflected:
-
-Installing dependencies:
+Installing dependencies for frontend:
 ```shell
 (sapp) $ cd sapp/ui/frontend && npm install
 ```
 
-Build static files and run the flask server:
+To interface sapp through the UI, we need to run both the frontend
+and backend simultaneously. In a production environment they run
+in the same port (because we serve the compilled version of frontend
+through the flask app in production). But in a development environment,
+the frontend runs in port 3000 and the backend runs in port 5000.
+
+Run the flask server and react app in development mode:
 ```shell
-(sapp) $ cd sapp/ui/frontend && npm run-script build
 (sapp) $ python3 -m sapp.cli server --debug
+(sapp) $ cd sapp/ui/frontend && npm run-script start
 ```
 
 ## FAQ
--- a/sapp/ui/frontend/src/index.js
+++ b/sapp/ui/frontend/src/index.js
@@ -27,8 +27,12 @@ import 'antd/dist/antd.css';
 const {Header, Content, Footer} = Layout;
 const {Text} = Typography;
 
+let uri = '/graphql';
+if (process.env.NODE_ENV === "development") {
+  uri = 'http://localhost:5000/graphql';
+}
 const client = new ApolloClient({
-  uri: '/graphql',
+  uri: uri,
   cache: new InMemoryCache(),
 });
 
--- a/.pyre_configuration
+++ b/.pyre_configuration
@@ -52,6 +52,9 @@
     {
       "site-package": "pyre_extensions"
     },
+    {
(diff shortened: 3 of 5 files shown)


abishekvashok:
Any idea as to why pyre can't detect flask-cors ?

0xedward:
> Any idea as to why pyre can't detect flask-cors ?

It's likely Pyre is failing, because the path of where packages are installed doesn't match where Pyre searches for those packages (unless there's a symlinks present and Pyre doesn't end up resolving them)

The log shows packages are installed in `/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/*`, but Pyre searches for those modules in `/home/runner/.local/lib/python3.9/site-packages/*` 

Probably we either need to update `.pyre_configuration` or update `pyre.yml` to run a virtual environment instead. @dkgi might know more about Pyre stuff

r0rshark:
> Any idea as to why pyre can't detect flask-cors ?

I think you should need to add it to the site-packages in https://github.com/facebook/sapp/blob/master/.pyre_configuration (same problem we run into when the type info was missing for mako.template :) )

abishekvashok:
Fixed it!

sapp/ui/frontend/src/index.js
```diff
@@ -27,9 +27,16 @@ import 'antd/dist/antd.css';
...
+let uri = '/graphql';
+if (process.env.NODE_ENV === "development") {
+  uri = 'http://localhost:5000/graphql';
+}
 const client = new ApolloClient({
-  uri: '/graphql',
+  uri: uri,
   cache: new InMemoryCache(),
+  fetchOptions: {
+    mode: 'no-cors',
```
r0rshark, line 38: Why do we need this? If it is only needed for development can we just add this option in the development and keep the production without it?

r0rshark (changes requested):

sapp/ui/frontend/src/index.js
abishekvashok: We don't need this!

r0rshark:
Nice, looks good to me now

README.md
```diff
@@ -259,27 +259,25 @@ $ source ~/.venvs/sapp/bin/activate
...
-If you make any changes to files under `sapp/ui/frontend/*`, you will need to run `npm install` once to install dependencies and `npm run-script build` each time you make changes before running the flask server to see the changes you made reflected:
-
-Installing dependencies:
+Installing dependencies for frontend:
 ```shell
 (sapp) $ cd sapp/ui/frontend && npm install
 ```
 
-Build static files and run the flask server:
+Run the flask server and react app in development mode:
```
0xedward, line 277: It seems like we need to run the flask server and `npm run-script start` simultaneously. Could you add a line in the docs to communicate that?

0xedward (changes requested):
This is a great idea! It would definitely help with the development workflow :)

Had a suggestion for the docs

README.md
abishekvashok: Sure

abishekvashok: @0xedward thanks. How does it look now 🤗?

README.md
```diff
@@ -259,27 +259,31 @@ $ source ~/.venvs/sapp/bin/activate
...
 ```shell
 (sapp) $ cd sapp/ui/frontend && npm install
 ```
 
-Build static files and run the flask server:
+To interface sapp through the UI, we need to run both the frontend
+and backend simultaneously. In a production environment they run
+in the same port (because we serve the compilled version of frontend
+through the flask app in production). But in a development environment,
+the frontend runs in port 3000 and the backend runs in port 5000.
```
0xedward, line 281:
```suggestion
To run SAPP with hot reloading of the Web UI, you need have the frontend and backend running simultaneously. In a production environment, they run in the same port. But in a development environment, the frontend runs in port 3000 and the backend runs in port 5000. You can indicate to SAPP to run in the development environment with the `debug` flag 
```

0xedward (changes requested):
Since the text will be in developer environment setup section, I had some wording suggestions

r0rshark:
> Since the text will be in developer environment setup section, I had some wording suggestions

I'll modify this myself in the imported diff :). I had to modify some stuff to make it work for our internal configuration :)

abishekvashok:
@r0rshark are you sure? The suggestion for me is a single button and a rebase away. But I don't wanna cause you trouble, if you have to reimport and do the work all over - either way works :)

r0rshark:
> @r0rshark are you sure? The suggestion for me is a single button and a rebase away. But I don't wanna cause you trouble, if you have to reimport and do the work all over - either way works :)

Yes, I just dealt with the internal issue and landed it :)

abishekvashok:
Thank you!

PR #37, opened 2021-08-03
Suppress sqlite3.ProgrammingError

abishekvashok:
Currently, sessions were created in threads and removed in the main
thread. This lead to sql alchemy complaining by emitting an
sqlite3.ProgrammingException. Moves local session removal to the same
thread in which it was created with the help of flask's teardown_request
callback and context stacks.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/42

### DIFF
--- a/sapp/ui/server.py
+++ b/sapp/ui/server.py
@@ -10,7 +10,9 @@ import os
 from typing import Optional
 
 import sqlalchemy
-from flask import Flask, send_from_directory
+
+# pyre-fixme[21]: `flask` has no attribute `_app_ctx_stack`
+from flask import Flask, send_from_directory, _app_ctx_stack
 from flask.wrappers import Response
 from flask_graphql import GraphQLView
 from pyre_extensions import none_throws
@@ -34,12 +36,11 @@ application = Flask(
 session: Optional[Session] = None
 
 
-@application.teardown_appcontext
+@application.teardown_request
 def shutdown_session(exception: Optional[Exception] = None) -> None:
-    local_session = session
-    if local_session is not None:
+    if session is not None:
         # pyre-fixme[16]: `Session` has no attribute `remove`.
-        local_session.remove()
+        session.remove()
 
 
 @application.route("/", defaults={"path": ""})
@@ -67,7 +68,11 @@ def start_server(
         echo=False,
         poolclass=None,
     )
-    session = scoped_session(sessionmaker(bind=engine))
+    session = scoped_session(
+        sessionmaker(bind=engine),
+        # pyre-fixme[16]: `flask` has no attribute _app_ctx_stack
+        scopefunc=_app_ctx_stack.__ident_func__,
+    )
     # pyre-fixme[16]: `Type` has no attribute `query`.
     models.Base.query = session.query_property()
     # We have additional tables for the UI that need to be created.


0xedward:
It seems like we are using [`scoped_session`](https://github.com/facebook/sapp/blob/HEAD/sapp/db.py#L87) instead. Since this exception is thrown for each transaction we create and commit, I wonder if the problem is actually related to how we are [managing our sessions](https://docs.sqlalchemy.org/en/12/orm/contextual.html?highlight=session%20management#using-thread-local-scope-with-web-applications) (e.g.  not calling `session.close()` after `session.commit()`, `Session.remove()` at the end of handling requests to flask, etc).

@abishekvashok is this something you could look into?

abishekvashok:
Yea I will look into this :)

abishekvashok:
@0xedward I tried session.close() after session.commit() but not working. The exception is still raised (to add to it, the exception is raised when we try to destroy the session, that is exit out of the flask app, after making a mutation)

r0rshark:
@0xedward  My guess is that we are creating the session in one thread: the cli application where the db is created https://github.com/facebook/sapp/blob/master/sapp/ui/server.py#L65 but then we are performing the mutation in a different thread which is the one spawn by the flask application run. We can try to validate this by printing the thread id in https://github.com/facebook/sapp/blob/master/sapp/ui/server.py#L65 and then in one of th GraphQL mutation and check if those thread id matches with the one in the exception:
``
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 123145560453120 and this is thread id 123145565708288.
``
@abishekvashok  can you check this?

abishekvashok:
> @abishekvashok can you check this?

Sure!

abishekvashok:
@r0rshark @0xedward  here are my debug findings. 

When a we start the flask server, the code before `application.run()` occurs in the main thread (https://github.com/facebook/sapp/blob/0dbae133a5e05ab6b6725183b9bfbc6ecc0ee897/sapp/ui/server.py#L90). For me that thread id is `4475055616` atm.

Hence, the session is created in the thread `4475055616`. When flask receives requests, I can see it using multiple threads. Sometimes a single thread can full fill all responses in the queue, or another thread can pick it up. The following log illustrates it:
```bash
2021-08-04 15:59:36,254 [INFO] Serving `static/js/main.9c9121d0.chunk.js`... [123145395937280]
2021-08-04 15:59:36,255 [INFO] Serving `static/css/main.6df5d19a.chunk.css`... [123145412726784]
2021-08-04 15:59:36,256 [INFO] Serving `static/js/2.dcdb68c9.chunk.js`... [123145429516288]
2021-08-04 15:59:36,256 [INFO] Found static resource. [123145395937280]
2021-08-04 15:59:36,256 [INFO] Found static resource. [123145412726784]
2021-08-04 15:59:36,256 [INFO] Found static resource. [123145429516288]
2021-08-04 15:59:36,380 [INFO] 127.0.0.1 - - [04/Aug/2021 15:59:36] "POST /graphql HTTP/1.1" 200 - [123145412726784]
2021-08-04 15:59:36,553 [INFO] 127.0.0.1 - - [04/Aug/2021 15:59:36] "POST /graphql HTTP/1.1" 200 - [123145395937280]
2021-08-04 15:59:36,784 [INFO] 127.0.0.1 - - [04/Aug/2021 15:59:36] "POST /graphql HTTP/1.1" 200 - [123145395937280]
2021-08-04 15:59:41,073 [INFO] 127.0.0.1 - - [04/Aug/2021 15:59:41] "POST /graphql HTTP/1.1" 200 - [123145395937280]
2021-08-04 16:00:27,765 [INFO] 127.0.0.1 - - [04/Aug/2021 16:00:27] "POST /graphql HTTP/1.1" 200 - [123145395937280]
2021-08-04 16:00:28,005 [INFO] 127.0.0.1 - - [04/Aug/2021 16:00:28] "POST /graphql HTTP/1.1" 200 - [123145395937280]
2021-08-04 16:00:37,988 [INFO] Deleting Zulip Triage copy 2 [123145379147776]
2021-08-04 16:00:37,991 [INFO] 127.0.0.1 - - [04/Aug/2021 16:00:37] "POST /graphql HTTP/1.1" 200 - [123145379147776]
2021-08-04 16:00:38,021 [INFO] 127.0.0.1 - - [04/Aug/2021 16:00:38] "POST /graphql HTTP/1.1" 200 - [123145379147776]
```

So, as you can see, mutations may run on different threads. As long as we are not overrunning the db with mutations, we should be safe to ignore the exception. One possible solution would be to initiate the connection on each request - but I don't think we would need that and I feel we can get away here :)

gbleaney (changes requested):
I don't think we want to just be suppressing this error. It looks like we're trying to use a `scoped_session` (which [seems to be the right idea according to SQL Alchemy docs](https://docs.sqlalchemy.org/en/14/orm/contextual.html)):
https://github.com/facebook/sapp/blob/d7698320276f5b696f145a202cfc012c210ff8e4/sapp/ui/server.py#L58-L70

The problem I see is that we're tying it to the life of the application, rather than the request, which is [what the SQL Alchemy docs propose](https://docs.sqlalchemy.org/en/14/orm/contextual.html#using-thread-local-scope-with-web-applications).:

https://github.com/facebook/sapp/blob/d7698320276f5b696f145a202cfc012c210ff8e4/sapp/ui/server.py#L38-L42

The docs also explicitly suggest using [Flask-SQLAlchemy](https://flask-sqlalchemy.palletsprojects.com/en/2.x/quickstart/) over the approach we're using.

So, before trying to land error suppression, can we evaluate changing our system to behave the "right" way (as defined by SQLAlchemy docs)? Switching to Flask-SQLAlchemy might be too big a change, but fixing up session lifetimes does seem pretty reasonable.

0xedward:
> The problem I see is that we're tying it to the life of the application, rather than the request, which is what the SQL Alchemy docs propose

It looks like [SQLAlchemy docs](https://docs.sqlalchemy.org/en/13/orm/contextual.html#using-custom-created-scopes) also provide some pseudocode for this. This [StackOverflow answer](https://stackoverflow.com/questions/37844704/flask-sqlalchemy-sessions-out-of-sync/37845889#37845889) also might be helpful

Maybe something roughly like the following in [sapp/ui/server.py](https://github.com/facebook/sapp/blob/d7698320276f5b696f145a202cfc012c210ff8e4/sapp/ui/server.py) would work?
```python
Session = scoped_session(sessionmaker(bind=some_engine), scopefunc=flask._app_ctx_stack.__ident_func__)

@app.teardown_request
def remove_session(req):
    Session.remove()
```

abishekvashok:
I am looking into this :) yes it occurs due to the way we are terminating the session at app teardown. I am trying some flask callbacks. teardown request for some reason occurs in the main thread. Will keep you all posted :)

abishekvashok:
Hey @0xedward the scopefunc in the scoped_session did the trick - now session are created and deleted in the same thread and the exception is no more! Wow 🥳

abishekvashok:
@0xedward @r0rshark @gbleaney done :)
Special thanks for mentioning scopefunc parameter and flask's application context stack @0xedward

abishekvashok:
Can affirm black is not failing due to my changes. I can fix it however :)

PR #38, opened 2021-08-05
Support regex matching for source and sink filters

abishekvashok:
Adds support for regex matching for source name and kind and sink name
and kind filters. The regex matching occurs in the frontend and no
modification is hence required in the backend.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/46

### DIFF
--- a/sapp/filter.py
+++ b/sapp/filter.py
@@ -10,14 +10,24 @@ from __future__ import annotations
 import json
 from json import JSONEncoder
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, TypedDict, Union
 
 from sqlalchemy import Column, String
 
 from .models import Base
 
 if TYPE_CHECKING:
-    from .ui.schema import FeatureCondition
+    from .ui.schema import FeatureCondition, MatchesIsField
+
+
+class NewMatchesIsAlias(TypedDict):
+    operation: str
+    value: List[str]
+
+
+# Previously the field contained values in form of a list, maintain Union for
+# backward compatibility.
+MatchesIsAlias = Union[NewMatchesIsAlias, List[str]]
 
 
 class FilterRecord(Base):
@@ -44,12 +54,12 @@ class Filter:
         )
         self.codes: List[int] = kwargs.get("codes", [])
         self.paths: List[str] = kwargs.get("paths", [])
-        self.callables: List[str] = kwargs.get("callables", [])
+        self.callables: MatchesIsAlias = kwargs.get("callables", [])
         self.statuses: List[str] = kwargs.get("statuses", [])
-        self.source_names: List[str] = kwargs.get("source_names", [])
-        self.source_kinds: List[str] = kwargs.get("source_kinds", [])
-        self.sink_names: List[str] = kwargs.get("sink_names", [])
-        self.sink_kinds: List[str] = kwargs.get("sink_kinds", [])
+        self.source_names: MatchesIsAlias = kwargs.get("source_names", [])
+        self.source_kinds: MatchesIsAlias = kwargs.get("source_kinds", [])
+        self.sink_names: MatchesIsAlias = kwargs.get("sink_names", [])
+        self.sink_kinds: MatchesIsAlias = kwargs.get("sink_kinds", [])
         self.traceLengthFromSources: Optional[List[int]] = kwargs.get(
             "traceLengthFromSources", None
         )
@@ -92,12 +102,12 @@ class Filter:
     def from_query(
         codes: List[int],
         paths: List[str],
-        callables: List[str],
+        callables: Optional[MatchesIsField],
         statuses: List[str],
-        source_names: List[str],
-        source_kinds: List[str],
-        sink_names: List[str],
-        sink_kinds: List[str],
+        source_names: Optional[MatchesIsField],
+        source_kinds: Optional[MatchesIsField],
+        sink_names: Optional[MatchesIsField],
+        sink_kinds: Optional[MatchesIsField],
         features: Optional[List[FeatureCondition]],
         min_trace_length_to_sinks: Optional[int],
         max_trace_length_to_sinks: Optional[int],
--- a/sapp/ui/filter_predicates.py
+++ b/sapp/ui/filter_predicates.py
@@ -5,8 +5,19 @@
 
 from __future__ import annotations
 
+import re
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Generic, List, Optional, Sequence, Set, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Generic,
+    List,
+    Optional,
+    Pattern,
+    Sequence,
+    Set,
+    TypeVar,
+    Union,
+)
 
 from sqlalchemy import Column
 from sqlalchemy.orm.query import Query
@@ -97,46 +108,42 @@ class HasAll(IssuePredicate):
         ]
 
 
-class HasAny(IssuePredicate):
-    def __init__(
-        self, parameter_list: Set[str], parameter_type: str = "features"
-    ) -> None:
-        self._parameter_list = parameter_list
-        self._parameter_type = parameter_type
+class Matches(IssuePredicate):
+    def __init__(self, regex: str, parameter_name: str) -> None:
+        self._regex: Pattern[str] = re.compile(regex)
+        self._parameter_name = parameter_name
+
+    def attribute_set(self, issue: IssueQueryResult) -> Set[str]:
+        attribute = issue._asdict()[self._parameter_name]
+        if isinstance(attribute, str):
+            return {attribute}
+        return set(attribute)
 
     def apply(self, issues: List[IssueQueryResult]) -> List[IssueQueryResult]:
-        if self._parameter_type == "features":
-            return [
-                issue
-                for issue in issues
-                if len(issue.features & self._parameter_list) > 0
-            ]
-        elif self._parameter_type == "source_names":
-            return [
-                issue
-                for issue in issues
-                if len(issue.source_names & self._parameter_list) > 0
-            ]
-        elif self._parameter_type == "source_kinds":
-            return [
-                issue
-                for issue in issues
-                if len(issue.source_kinds & self._parameter_list) > 0
-            ]
-        elif self._parameter_type == "sink_names":
-            return [
-                issue
-                for issue in issues
-                if len(issue.sink_names & self._parameter_list) > 0
-            ]
-        elif self._parameter_type == "sink_kinds":
-            return [
-                issue
-                for issue in issues
-                if len(issue.sink_kinds & self._parameter_list) > 0
-            ]
-        else:
-            return issues
+        return [
+            issue
+            for issue in issues
+            if any(map(self._regex.match, self.attribute_set(issue)))
+        ]
+
+
+class HasAny(IssuePredicate):
+    def __init__(self, parameter_list: Set[str], parameter_name: str) -> None:
+        self._parameter_list = parameter_list
+        self._parameter_name = parameter_name
+
+    def attribute_set(self, issue: IssueQueryResult) -> Set[str]:
+        attribute = issue._asdict()[self._parameter_name]
+        if isinstance(attribute, str):
+            return {attribute}
+        return set(attribute)
+
+    def apply(self, issues: List[IssueQueryResult]) -> List[IssueQueryResult]:
+        return [
+            issue
+            for issue in issues
+            if not self.attribute_set(issue).isdisjoint(self._parameter_list)
+        ]
 
 
 class HasNone(IssuePredicate):
--- a/sapp/ui/frontend/src/Filter.js
+++ b/sapp/ui/frontend/src/Filter.js
@@ -195,31 +195,37 @@ const MatchesIsField = (
     setCurrentFilter: FilterDescription => void,
   }>,
 ): React$Node => {
-  const [mode, setMode] = useState('is');
-  const [inputValue, setInputValue] = useState(null);
+  const [inputValue, setInputValue] = useState('');
+  const [selectValue, setSelectValue] = useState([]);
+  const [matchedValues, setMatchedValues] = useState([]);
+
+  const triggerChange = (operation: string, value: mixed = undefined): void => {
+    let values = {};
+    if(value === undefined) {
+      value = (operation === 'is') ? selectValue : inputValue;
+    }
+    values[props.parameterName] = {
+      operation: operation,
+      value: value,
+    };
+    props.setCurrentFilter({...props.currentFilter, ...values});
+  }
+
+  const selectedOperation =
+    props.currentFilter[props.parameterName]?.operation || 'is';
 
   return (
     <Row gutter={gutter}>
       <Col span={6}>
         <Select
           options={[{value: 'is'}, {value: 'matches'}]}
-          value={mode}
-          onChange={setMode}
+          value={selectedOperation}
+          onChange={triggerChange}
           style={{width: '100%'}}
         />
       </Col>
       <Col span={16}>
-        {mode === 'is' ? (
-          <Select
-            mode="multiple"
-            value={props.currentFilter[props.parameterName]}
-            options={props.allOptions.map(value => ({value}))}
-            style={{width: '100%'}}
-            onChange={value =>
-              props.setCurrentFilter({...props.currentFilter, value})
-            }
-          />
-        ): (
+        {selectedOperation === 'matches' ? (
           <Input
             placeholder="regular expression"
             style={{width: '100%'}}
@@ -230,19 +236,29 @@ const MatchesIsField = (
               const option_values = props.allOptions.filter(option =>
                   option.match(value),
               );
-              let values = {};
-              values[props.parameterName] = option_values;
-              props.setCurrentFilter({...props.currentFilter, ...values});
+              setMatchedValues(option_values);
+              triggerChange('matches', value);
             }}
             suffix={
               <Tooltip
-                title={(props.currentFilter[props.parameterName] || []).join('\n')}
+                title={(matchedValues || []).join('\n')}
                 placement="bottom">
                 <Text type="secondary" size="small">
-                  {(props.currentFilter[props.parameterName] || []).length}
+                  {(matchedValues || []).length}
                 </Text>
               </Tooltip>
             }
+          />
+        ): (
+          <Select
+            mode="multiple"
+            value={selectValue}
+            options={props.allOptions.map(value => ({value}))}
+            style={{width: '100%'}}
+            onChange={value => {
+              setSelectValue(value);
+              triggerChange('is', value);
+            }}
           />
         )}
       </Col>
--- a/sapp/ui/frontend/src/Filter.js
+++ b/sapp/ui/frontend/src/Filter.js
@@ -187,6 +187,69 @@ const Paths = (
   );
 };
 
+const MatchesIsField = (
+  props: $ReadOnly<{
+    allOptions: $ReadOnlyList<string>,
+    parameterName: string,
+    currentFilter: FilterDescription,
+    setCurrentFilter: FilterDescription => void,
+  }>,
+): React$Node => {
+  const [mode, setMode] = useState('is');
+  const [inputValue, setInputValue] = useState(null);
+
+  return (
+    <Row gutter={gutter}>
+      <Col span={6}>
+        <Select
+          options={[{value: 'is'}, {value: 'matches'}]}
+          value={mode}
+          onChange={setMode}
+          style={{width: '100%'}}
+        />
+      </Col>
+      <Col span={16}>
+        {mode === 'is' ? (
+          <Select
+            mode="multiple"
+            value={props.currentFilter[props.parameterName]}
+            options={props.allOptions.map(value => ({value}))}
+            style={{width: '100%'}}
+            onChange={value =>
+              props.setCurrentFilter({...props.currentFilter, value})
+            }
+          />
+        ): (
+          <Input
+            placeholder="regular expression"
+            style={{width: '100%'}}
+            value={inputValue}
+            onChange={event => {
+              const value = event.target.value;
+              setInputValue(value);
+              const option_values = props.allOptions.filter(option =>
+                  option.match(value),
+              );
+              let values = {};
+              values[props.parameterName] = option_values;
+              props.setCurrentFilter({...props.currentFilter, ...values});
+            }}
+            suffix={
+              <Tooltip
+                title={(props.currentFilter[props.parameterName] || []).join('\n')}
+                placement="bottom">
+                <Text type="secondary" size="small">
+                  {(props.currentFilter[props.parameterName] || []).length}
+                </Text>
+              </Tooltip>
+            }
+          />
+        )}
+      </Col>
+    </Row>
+  );
+};
+
 const SourceNames = (
   props: $ReadOnly<{
     currentFilter: FilterDescription,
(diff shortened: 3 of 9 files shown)


abishekvashok:
After the changes:

Screenshot 2021-08-05 at 11 22 07 AM

sapp/ui/frontend/src/Filter.js
```diff
@@ -211,17 +214,50 @@ const SourceNames = (
...
+              value={props.currentFilter.source_names}
+              options={allSourceNames.map(value => ({value}))}
+              style={{width: '100%'}}
+              onChange={source_names =>
+                props.setCurrentFilter({...props.currentFilter, source_names})
+              }
+            />
+          ): (
+            <Input
+              placeholder="regular experssion"
```
r0rshark, line 238:
```suggestion
              placeholder="regular expression"
```

r0rshark: Can you fix also the other instances?

r0rshark (changes requested):
Besides the typo it looks good to me

sapp/ui/frontend/src/Filter.js
abishekvashok: Sure

abishekvashok: @r0rshark done

r0rshark:
@abishekvashok  can you check why black is complaining?

abishekvashok:
@r0rshark can affirm black is not failing due to the changes - I haven't touched any py file in the diff. I can fix or supress it however :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -251,17 +290,50 @@ const SourceKinds = (
...
+                  title={(props.currentFilter.source_kinds || []).join('\n')}
+                  placement="bottom">
+                  <Text type="secondary" size="small">
+                    {(props.currentFilter.source_kinds || []).length}
+                  </Text>
+                </Tooltip>
+              }
+            />
+          )}
+        </Col>
```
dkgi, line 336: Looks like we repeat this pattern a couple of times. Do you think this could be factored out into its own component?

dkgi (changes requested):
This look good to me. Back for question.

sapp/ui/frontend/src/Filter.js
abishekvashok: I will try!

abishekvashok:
@dkgi done :)

r0rshark:
Looks good I wonder why the ESLint CI seems to have run on the previous version and not on the last one

abishekvashok:
@r0rshark should be good now :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -187,12 +187,76 @@ const Paths = (
   );
 };
 
+const MatchesIsField = (
+  props: $ReadOnly<{
+    allOptions: mixed,
```
dkgi, line 192: Should this be `$ReadOnlyList<string>`?

sapp/ui/frontend/src/Filter.js
```diff
@@ -187,12 +187,76 @@ const Paths = (
...
+  );
+};
+
 const SourceNames = (
   props: $ReadOnly<{
     currentFilter: FilterDescription,
     setCurrentFilter: FilterDescription => void,
   }>,
 ): React$Node => {
+
```
dkgi, line 259: Is the change here coming from the auto-formatter?

dkgi (changes requested):
Looks good to me. Minor nit.

sapp/ui/frontend/src/Filter.js
abishekvashok: Yes, kinda added a wierd vim plugin. Will remove it.

sapp/ui/frontend/src/Filter.js
abishekvashok: That's a much better type description!

gbleaney:
Does client side regex matching really make sense here? Shouldn't this be something we push down to the DB query? I don't want us to have to return the entire DB to the client just to find that one issues with a unique sink

abishekvashok:
> Does client side regex matching really make sense here? Shouldn't this be something we push down to the DB query? I don't want us to have to return the entire DB to the client just to find that one issues with a unique sink

We are getting a list of sources relating to a run already to show options, just like every other option to filtering. We are matching with that set just like we do for callables!

abishekvashok:
@dkgi the requested changes have been made. Thank you for your time :)

r0rshark:
> ex matching really make sense here? Shouldn't this be something we push down to the DB query? I don't want us to have to return the entire DB to the client just to find that one issues with a unique sink

It's slightly different than client side filtering. The way it works is:
0. We fetch the possible "sink kind" values e.g. "ReturnedToUser", "XSS"
1. Clientside  we do regex matching between the input field and the possible "sink kinds" e.g. with Input value "Retur" -> will get a list containing  "ReturnedToUser"
2. We send a graphQL query is sink_kinds: ["ReturnedToUser"]

I agree this is still not optimal since we need to return to the client the list of all the possible "Sink Kinds"/Names but it's better than returning the entire db :)
This approach was initially implemented by @dkgi who may have more context 
Screenshot 2021-08-09 at 13 57 41

r0rshark:
> ex matching really make sense here? Shouldn't this be something we push down to the DB query? I don't want us to have to return the entire DB to the client just to find that one issues with a unique sink

It's slightly different than client side filtering. The way it works is:
0. We fetch the possible "sink kind" values e.g. "ReturnedToUser", "XSS"
1. Clientside  we do regex matching between the input field and the possible "sink kinds" e.g. with Input value "Retur" -> will get a list containing  "ReturnedToUser"
2. We send a graphQL query is sink_kinds: ["ReturnedToUser"]

I agree this is still not optimal since we need to return to the client the list of all the possible "Sink Kinds"/Names but it's better than returning the entire db :)
This approach was initially implemented by @dkgi who may have more context 
Screenshot 2021-08-09 at 13 57 41

0xedward:
> 0. We fetch the possible "sink kind" values e.g. "ReturnedToUser", "XSS"
>1. Clientside we do regex matching between the input field and the possible "sink kinds" e.g. with Input value "Retur" -> will get a list containing "ReturnedToUser"
>2. We send a graphQL query is sink_kinds: ["ReturnedToUser"]

> Does client side regex matching really make sense here? Shouldn't this be something we push down to the DB query? I don't want us to have to return the entire DB to the client just to find that one issues with a unique sink

It seems possible to also do this by:
1. passing the regex from the UI to the flask server with a graphql query
2. extending SQLAlchemy `String` class (e.g. https://gist.github.com/Xion/204ddbd020f1a4275a53) 
3. adding something like `where_regex_matches` in [`Instance`](https://github.com/facebook/sapp/blob/4adba9589f3ae28dd265c5597fbf8b99d39e75a4/sapp/ui/issues.py#L223) to build the DB query
4. run the DB query built in step 3 and return the list of all sources/sinks that match the regex

dkgi:
I think we originally did this on the client side because it was cheap enough (even sending all possible function names to the client wasn't a big deal). It also allows us to provide live feedback to the user what their current regex is matching which is nice. Downside here is that it's hard to share the logic, and could potentially suck if we're sending back a large set to the query...

Don't have strong opinions myself.

abishekvashok:
I can change this if we want to do it the other way, but as @dkgi said, we already have a list of source names (and every other field with us on the frontend to show as suggestions). We pass the needed ones to the backend in the form of a `filterDescription` to specify a filter. A regex matching and filtering by regex hence shouldn't be a big deal. The passed filters would look like the same when passed without the regex. But, I can change this if we would like to make a switch as @0xedward mentions :)

gbleaney:
Ahh, I didn't realize we were doing *all* regex filtering client side. I get that it's not a perf problem (right now), but it still does present other problems. My biggest concern is: doesn't this make CLI filtering incompatible with UI filtering? IE. I can create a filter in the UI and get it perfect using RegEx components, but be unable to run it from the command line. That's a problem, because we assume that when people are using Pysa in CI, they're going to want to run the command line filtering rather than start up a whole webserver and headless browser in their CI.

Assuming that that concern is correct, @r0rshark and @abishekvashok, I'm pretty strongly in favour of switching to server side evaluation of the regex. I've got no problem with *also* offering live matching on the client side (assuming you can ensure you've got compatible regex engines running in both places) to make the UX better, but the canonical filtering should be done in the backend so CLI and UI filters are compatible.

abishekvashok:
@gbleaney nice thought!
Currently, backend didn't have regex support - so without the sapp frontend regex isn't available via cli.
In the long term, I guess the changes pointed makes more sense, so I moved the regex matching to backend (also migrated existing filter parameters that had regex matching [callables] and their tests).

r0rshark (changes requested):
Thanks for working on this! From my understanding with these changes we would be able to support creating a filter with regex filtering in the UI and using it in the commandline to filter issues. Can you validate this?
I just have a concern in the way we store the regex filtering information in the filter because having the type of match as the first element in the list look a bit hackish

sapp/ui/filter_predicates.py
```diff
@@ -103,40 +114,36 @@ def __init__(
     ) -> None:
         self._parameter_list = parameter_list
         self._parameter_type = parameter_type
+        self._use_regex = (
+            True
+            if len(parameter_list) == 2 and "__regex_match__" in parameter_list
```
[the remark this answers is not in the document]
abishekvashok: Yeah, I added `"__regex_match__"` as the first element as I was trying a way to keep the existing filter layout! (Backward compatibility). I can change it to the mentioned layout, but I don't know whether existing filters will work :/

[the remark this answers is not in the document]
gbleaney: A good middle ground here is that you can check the type that you received. If you received a list, treat it as an `equals` operation and emit a warning saying that the user should upgrade their filter to the new format. If you receive a dict, parse as normal.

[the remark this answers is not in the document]
r0rshark: I agree with Graham, let's try this approach :)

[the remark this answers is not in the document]
abishekvashok: Done :)

abishekvashok:
Fixing tests now :)

abishekvashok:
@r0rshark @gbleaney thanks.

> From my understanding with these changes we would be able to support creating a filter with regex filtering in the UI and using it in the commandline to filter issues. Can you validate this?

As requested here's a testing of the backend.

https://user-images.githubusercontent.com/8947010/129366782-166ed985-2daf-4479-b27b-f92d4f2aeb46.mov

Also, I found out that we would need to export a filter created in the backend as a json file, and then use the json file to query. Potentially, I would love to make a feature request to be able to filter with just filter name on the backend (^^)

0xedward:
> Also, I found out that we would need to export a filter created in the backend as a json file, and then use the json file to query. Potentially, I would love to make a feature request to be able to filter with just filter name on the backend (^^)

@abishekvashok The current `filter issues` command was built for the upcoming SARIF support and was intended to be used in CI. 

I think that would be a good feature to add to our IPython interface ([`interactive.py`](https://github.com/facebook/sapp/blob/main/sapp/ui/interactive.py)) :)

sapp/ui/frontend/src/Filter.js
```diff
@@ -187,6 +187,86 @@ const Paths = (
...
+        {selectedOperation === 'matches' ? (
+          <Input
+            placeholder="regular expression"
+            style={{width: '100%'}}
+            value={inputValue}
+            onChange={event => {
+              const value = event.target.value;
+              setInputValue(value);
+              const option_values = props.allOptions.filter(option =>
+                  option.match(value),
```
gbleaney, line 238: Do Python and JavaScript use the same Regex engines? If not, trying to run the same regex on both might end up with confusing subtleties where frontend results don't match backend results

sapp/ui/issues.py
```diff
@@ -462,26 +476,100 @@ def where_filter(self, filter_instance: Filter) -> "Instance":
             .where_status_is_any_of(filter_instance.statuses)
         )
 
+        if filter_instance.callables is not None and len(filter_instance.callables) > 0:
```
gbleaney, line 479: Both `None` and empty collections are falsey, so this is more concise
```suggestion
        if filter_instance.callables:
```

sapp/ui/issues.py
```diff
@@ -462,26 +476,100 @@ def where_filter(self, filter_instance: Filter) -> "Instance":
...
+            if type(filter_instance.callables) is not list:
```
gbleaney, line 480: I'm confused my this. Above, `callables` is typed as a `List[Dict[...`, and before it was a `List[str]`. Shouldn't this branch never be hit? Based on the code below, it looks like you expect it to be a `Dict`. If that's the case, I think you need to update the typing on `Filter`

sapp/ui/issues.py
```diff
@@ -462,26 +476,100 @@ def where_filter(self, filter_instance: Filter) -> "Instance":
...
+                if filter_instance.callables.get("operation") == "matches":
+                    builder = builder.where_callables_matches(
+                        filter_instance.callables.get("value", [""])[0]
+                    )
+                elif len(filter_instance.callables.get("value", [])) > 0:
+                    builder = builder.where_callables_is_any_of(
+                        filter_instance.callables.get("value", ["%"])
```
gbleaney, line 487: We have a default value of `["%"]` here, but nowhere else. I get that you're propagating the default value of `%` that you deleted from the `Query` object below, but I think that default was needed due to the unconditional `where_callables_is_any_of` that was being done above. Now that you've removed that, we'll only run the `where_callables_is_any_of` check if we actually have something to filter on, so I don't think the default is needed any more.

sapp/ui/tests/issues_test.py
```diff
@@ -150,8 +150,10 @@ def testWhereCallables(self) -> None:
             builder = Instance(session, latest_run_id)
             issue_ids = {
                 int(issue.issue_instance_id)
-                for issue in builder.where_callables_is_any_of(["%function3"])
-                .where_callables_is_any_of(["%sub%"])
+                for issue in builder.where_callables_is_any_of(
+                    ["__regex_match__", ".*function3"]
+                )
+                .where_callables_is_any_of(["__regex_match__", ".*sub.*"])
```
gbleaney, line 156: Out of date test?

sapp/ui/tests/issues_test.py
```diff
@@ -150,8 +150,10 @@ def testWhereCallables(self) -> None:
             builder = Instance(session, latest_run_id)
             issue_ids = {
                 int(issue.issue_instance_id)
-                for issue in builder.where_callables_is_any_of(["%function3"])
```
gbleaney, line 153: Could you please add tests for the source and sink filtering you added in this PR?

gbleaney (changes requested):
Thanks for the big refactor! Looks pretty good overall, just some small concerns to address

sapp/ui/issues.py
```diff
@@ -462,26 +476,100 @@ def where_filter(self, filter_instance: Filter) -> "Instance":
             .where_status_is_any_of(filter_instance.statuses)
         )
 
+        if filter_instance.callables is not None and len(filter_instance.callables) > 0:
+            if type(filter_instance.callables) is not list:
+                if filter_instance.callables.get("operation") == "matches":
+                    builder = builder.where_callables_matches(
+                        filter_instance.callables.get("value", [""])[0]
+                    )
+                elif len(filter_instance.callables.get("value", [])) > 0:
```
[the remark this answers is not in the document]
abishekvashok: I guess that makes the else hit on filters with empty callables such as:
```json
{
operation: "is",
value: ""
}
```
and give out no issues! Is that intended?

sapp/ui/issues.py
abishekvashok: You are right! I was being too defensive (code) there!

sapp/ui/filter_predicates.py
```diff
@@ -97,46 +108,42 @@ def apply(self, issues: List[IssueQueryResult]) -> List[IssueQueryResult]:
...
+        attribute = issue._asdict()[self._parameter_name]
+        if type(attribute) is str:
+            return {attribute}
+        return set(attribute)
+
+    def apply(self, issues: List[IssueQueryResult]) -> List[IssueQueryResult]:
+        return [
+            issue
+            for issue in issues
+            if len(list(filter(self._regex.match, self.attribute_set(issue)))) > 0
```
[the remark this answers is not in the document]
abishekvashok: Really insightful! Thanks, it makes me happy as well :)

sapp/ui/frontend/src/Filter.js
abishekvashok: I never thought we would have large differences, but it seems the case! So I updated the backend to use a library called js_regex!

sapp/ui/issues.py
abishekvashok: It was actually to be `Dict`. Updated the typing on Filter :)

sapp/ui/tests/issues_test.py
abishekvashok: Yeah :cheeky: missed that! Sorry.

sapp/ui/tests/issues_test.py
abishekvashok: We lacked tests for sources and sinks, added them :)

abishekvashok:
@gbleaney thank you for the in-depth review. As someone who enjoys tiny optimisations, loved the suggestions and enjoyed learning them :)

Fixed everything as suggested except one - left a comment just to make sure :)
Again, thanks for going in-depth 😁

requirements.txt
```diff
@@ -6,6 +6,7 @@ flask-cors
 graphene==2.1.3
 graphene_sqlalchemy
 ipython
+js-regex
```
0xedward, line 9:
```suggestion

```

.pyre_configuration
```diff
@@ -28,6 +28,9 @@
     {
       "site-package": "IPython"
     },
+    {
+      "site-package": "js_regex"
+    },
```
0xedward, line 33:
```suggestion
```

0xedward (changes requested):
LGTM :) I had a suggestion to remove the `js-regex` package, since it seems like it isn't being used.

abishekvashok:
@0xedward sorry I don't how I missed adding a file, but js-regex is used, it's me messing up - sorry :/

sapp/ui/filter_predicates.py
```diff
@@ -5,8 +5,19 @@
 
 from __future__ import annotations
 
+import js_regex
```
arthaud, line 8: nitpick: since this is not in the standard library, I would put it in the next import block (next to sqlalchemy stuff)

sapp/ui/filter_predicates.py
```diff
@@ -97,46 +108,42 @@ def apply(self, issues: List[IssueQueryResult]) -> List[IssueQueryResult]:
...
 
 
+class Matches(IssuePredicate):
+    def __init__(self, regex: str, parameter_name: str) -> None:
+        self._regex: Pattern[str] = js_regex.compile(regex)
+        self._parameter_name = parameter_name
+
+    def attribute_set(self, issue: IssueQueryResult) -> Set[str]:
+        attribute = issue._asdict()[self._parameter_name]
+        if type(attribute) is str:
```
arthaud, line 118: Here and other places: I would prefer `isinstance(attribute, str)` over `type(attribute) is str`.

sapp/ui/issues.py
arthaud: ditto: would use `isinstance`

sapp/filter.py
```diff
@@ -17,7 +17,9 @@
 from .models import Base
 
 if TYPE_CHECKING:
-    from .ui.schema import FeatureCondition
+    from .ui.schema import FeatureCondition, MatchesIsField
+
+MatchesIsAlias = Union[Dict[str, Union[str, List[str]]], List[str]]
```
arthaud, line 22: I think we could use a `TypedDict` here, to better describe what that `Dict[str, Union[str, List[str]]]` looks like.

sapp/ui/issues.py
```diff
@@ -488,26 +502,95 @@ def where_filter(self, filter_instance: Filter) -> "Instance":
...
-        ):
-            builder = builder.where_sink_name_is_any_of(filter_instance.sink_names)
-        if (
-            filter_instance.sink_kinds is not None
-            and len(filter_instance.sink_kinds) > 0
-        ):
-            builder = builder.where_sink_kind_is_any_of(filter_instance.sink_kinds)
+        if filter_instance.callables:
+            if type(filter_instance.callables) is not list:
+                # pyre-fixme[16]: `List` has no attribute 'get'
```
arthaud, line 507: Why is that necessary? pyre should be able to infer that we have a `Dict` here.

sapp/ui/issues.py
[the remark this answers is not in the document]
arthaud: I would have a check `elif filter_instance.callables.get("operation") == "is":` and a `else` that throws an error.

sapp/ui/issues.py
```diff
@@ -488,26 +502,95 @@ def where_filter(self, filter_instance: Filter) -> "Instance":
...
+                    )
+                elif len(filter_instance.callables.get("value", [])) > 0:
+                    builder = builder.where_callables_is_any_of(
+                        filter_instance.callables.get("value")
+                    )
+            else:
+                # Backward compatibility
+                # pyre-fixme[6]: Expected `List[str]` for 1st positional only
+                # parameter but got
+                # `Union[Dict[str, Union[List[str], str]], List[str]]`
```
arthaud, line 520: ditto, surprised pyre cannot infer that we have a list here. Maybe because you use `type(..) is list` instead of `isinstance()`?

arthaud (changes requested):

sapp/ui/filter_predicates.py
abishekvashok: Yep moving it :)

sapp/ui/filter_predicates.py
abishekvashok: Sure

sapp/ui/issues.py
abishekvashok: Migrated to `isinstance()`. Still pyre fails!

sapp/ui/issues.py
abishekvashok: No it  isn't.

sapp/filter.py
abishekvashok: Done

abishekvashok:
@arthaud thanks for your time reviewing :) I tried all your suggestions and pushed everything except for removing the pyre-check fixme comments. Pyre for some reason can't infer the type. Maybe I am missing something, but not sure!

sapp/ui/filter_predicates.py
```diff
@@ -6,8 +6,19 @@
...
+    List,
+    Optional,
+    Pattern,
+    Sequence,
+    Set,
+    TypeVar,
+    Union,
+)
+
+import js_regex
```
r0rshark, line 21: @abishekvashok Why do we need to import js_regex can't we use the regular python regexes? js_regex https://pypi.org/project/js-regex/ looks like it simply support the javascript regex syntax which is not really something we care about. Moreover this adds a bit of maintenance burden since we would need to support this dependency  internally. Can we use normal python regex instead :)

r0rshark (changes requested):

sapp/ui/filter_predicates.py
abishekvashok: @r0rshark yes we can. But there are differences between the regex engines used by javascript and python. I switched to js-regex following this comment, https://github.com/facebook/sapp/pull/38#discussion_r690458934

But yeah, I will make the change and remove `js-regex` if we don't care much about the numbers in the frontend :)

abishekvashok:
@r0rshark done :)

abishekvashok:
@r0rshark done :)

PR #39, opened 2021-08-06
Adds optional pipeline to add features to issues

abishekvashok:
Adds optional pipeline to add features to all issues using the sapp
analyze command. All features provided using the --add-feature
flag/option shall be added to all existing issues associated with the
run.

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/47


abishekvashok:
Testing steps done:
Adding two features (one new, other existing) to be associated with all issues:
```bash
python3 -m sapp.cli analyze taint-output.json --add-feature feature:custom-demo-feature --add-feature always-type:scalar
```

Result:
Screenshot 2021-08-06 at 1 22 50 PM

abishekvashok:
The issues raised by flake8 and black aren't related to this PR. Since, it is persisting, I thought I would fix them for easier work flow for everyone. Hence made a PR: #40

r0rshark (changes requested):
Thanks! After reviewing the logic I think it make more sense to only affect the imported issues not all the one currently stored

sapp/pipeline/add_features.py
```diff
@@ -0,0 +1,103 @@
...
+    querying all issue instances, and creating rows in the table
+    IssueInstanceSharedTextAssoc."""
+
+    def __init__(self, database: DB, features: List[str]) -> None:
+        super().__init__()
+        self.database = database
+        self.features: Set[str] = set(features)
+
+    def run(self, input: RunSummary, summary: Summary) -> Tuple[RunSummary, Summary]:
+        with self.database.make_session() as session:
```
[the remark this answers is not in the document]
abishekvashok: I guess we are selecting only modifying those issue instances which are imported in the current run, 
`.filter(IssueInstance.run_id == input.id)` (also, pipeline is called after issues are imported to the db)

abishekvashok:
@r0rshark pushed the pipeline up the order as discussed. Now it's short and crisp :) (changes screenshot remains the same)

sapp/pipeline/add_features.py
```diff
@@ -0,0 +1,35 @@
...
+        # pyre-fixme[6]: Expected `typing.Iterable[Variable[_T]]` for `set.__init__`
+        # but got `Optional[List[str]]`
+        self.features: Set[str] = set(features)
+
+    def run(self, input: DictEntries, summary: Summary) -> Tuple[DictEntries, Summary]:
+        if len(self.features) > 0:
+            log.info("Attaching provided features")
+        # pyre-fixme[6]: Expected `typing.Sized` for 1st positional only parameter
+        # to call `len` but got `typing.Iterable[sapp.pipeline.ParseIssueTuple]`
+        for i in range(len(input["issues"])):
```
gbleaney, line 30: You can avoid all the lint errors and using the internal `_replace` API with something like this:
```suggestion
        for issue in input["issues"]:
            issue.feature.append(self.features.difference(set(issue.feature)))
```

sapp/pipeline/add_features.py
```diff
@@ -0,0 +1,35 @@
...
+class AddFeatures(PipelineStep[DictEntries, DictEntries]):
+    """Optional pipeline that attaches given features to all issues.
+    If the features list is empty, the input is simply passed without
+    change as the output"""
+
+    def __init__(self, features: Optional[List[str]]) -> None:
+        super().__init__()
+        # pyre-fixme[6]: Expected `typing.Iterable[Variable[_T]]` for `set.__init__`
+        # but got `Optional[List[str]]`
+        self.features: Set[str] = set(features)
```
gbleaney, line 23: We definitely shouldn't be suppressing an error here. Either `features` can't be `None`, and we should drop the `Optional` annotation, or it can be and this code will throw an exception when you do `set(None)`. Handling the optional like this
```suggestion
        self.features: Set[str] = set(features || [])
```

gbleaney (changes requested):

sapp/pipeline/add_features.py
abishekvashok: I tried that but the element in the list wasn't changing for some reason :/

abishekvashok: Another way of modification works :)

sapp/pipeline/add_features.py
abishekvashok: Used or and it seems working :)

abishekvashok:
@gbleaney thanks for the comments. Fixed all pyre bugs with your suggestions (I had to use `input["issues"] = [issue ...]` over `for issue in input["issue"]` though! I hope you don't mind that!

PR #40, opened 2021-08-06
[Chore]: Fix issues raised by flake8 and black

abishekvashok:
Fixes issues raised by flake8 (unused imports) and black code linting errors.
These errors were persisting for quite some time, so I though I would just
fix them :)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,7 +20,7 @@ jobs:
         run: pip install black==21.4b2
 
       - name: Run black
-        run: black --check --diff .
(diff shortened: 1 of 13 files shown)


r0rshark:
Thanks, looks like black need another minor fix

abishekvashok:
@r0rshark done :)

r0rshark:
@abishekvashok  besides the fix can we also exclude the pyi files from the checking similarly to what we did in pyre-check. It looks like our internal CI is not caring about those (we can keep the fix though).

abishekvashok:
Not sure if it's the way we did in pyre-check, but this should do it :)

PR #41, opened 2021-08-10
Add Jest tests to test ReactJS frontend

abishekvashok:
Adds Jest testing platform and basic tests to make sure Issues, Runs,
and Filters are rendering. Uses Apollo MockedProvider to produce mocked
responses so that tests can run without the flask backend.

Adds two tests for issues and runs: to check whether issues along with
filters or runs are being rendered when provided with a mock response
(first loading is checked, and then rendering).

Adds 4 snapshots - issues rendered, issues loading, runs rendered, runs
loading which will be checked against in the future to make sure we get
the intended results.

Adds a Github Action as well to enable these "ui-tests".

Fixes: https://github.com/MLH-Fellowship/pyre-check/issues/48

### DIFF
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -1,0 +1,22 @@
+name: jest
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+  workflow_dispatch:
+
+jobs:
+  jest:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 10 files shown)


abishekvashok:
Screenshot 2021-08-10 at 5 34 12 PM

NB: I am kinda perplexed as to how the issues snapshot looks.

abishekvashok:
> NB: I am kinda perplexed as to how the issues snapshot looks.

I was missing two ticks because we had two mock queries! Now it loads really well. Added additional tests as well :)

Screenshot 2021-08-12 at 12 46 43 PM

PR #42, opened 2021-08-12
Adds minimum width to status dropdown

abishekvashok:
Tiny nit adding minimum width to status dropdown so that the options
would never be displayed in a truncated fashion when the option with the
least length is selected.

Fixes part of https://github.com/MLH-Fellowship/pyre-check/issues/49


abishekvashok:
Before:
Screenshot 2021-08-12 at 4 57 52 PM

After:
Screenshot 2021-08-12 at 4 56 33 PM

PR #43, opened 2021-08-12
Surface first seen attribute of issue in UI

abishekvashok:
Surface first_seen attribute of Issue stored in the db in the UI. Also
fixes a minor margin problem (nit).

### DIFF
--- a/sapp/ui/frontend/src/index.css
+++ b/sapp/ui/frontend/src/index.css
@@ -17,6 +17,7 @@ body {
 code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
     monospace;
+  margin: 0 !important;
 }
 
 .header {
--- a/sapp/ui/frontend/src/Issue.js
+++ b/sapp/ui/frontend/src/Issue.js
@@ -138,6 +138,7 @@ export type IssueDescription = {
   features: $ReadOnlyArray<string>,
   min_trace_length_to_sources: number,
   min_trace_length_to_sinks: number,
+  first_seen: string,
   is_new_issue: boolean,
 };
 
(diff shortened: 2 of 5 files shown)


abishekvashok:
Before:
Screenshot 2021-08-12 at 5 16 24 PM

After:
Screenshot 2021-08-12 at 5 16 47 PM


We've just added the existing something existing in the db to the UI. We can probably implement filter controls for this - if so what would be preferred? Filter with data or filter with date and time (we can do possibly - before, after, and intervals for both) or would something like that be too fancy?

sapp/ui/frontend/src/index.css
```diff
@@ -17,6 +17,7 @@ body {
 code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
     monospace;
+  margin: 0 !important;
```
gbleaney, line 20: I know very little about CSS, but I know that `!important` is usually an anti-pattern. Why is this needed?

gbleaney (commented):
Diff LGTM, other than a small concerns about the `!important`

> We can probably implement filter controls for this - if so what would be preferred? Filter with data or filter with date and time (we can do possibly - before, after, and intervals for both) or would something like that be too fancy?

Internally we do filtering with optional before/after dates:
Screen Shot 2021-08-12 at 2 20 27 PM

Something like that would be reasonable here too. Definitely doesn't have to be a part of this PR though. I'll let you and @r0rshark  prioritize that feature request together.

gbleaney (changes requested):
.

r0rshark:
Let's also address the failing tests :)

sapp/ui/frontend/src/index.css
abishekvashok: if we don't set it to `!important`, antd element styles rendered after index.css renders, would override the margin and set a 0.2em margin. To prevent that overriding and enforcing of margin, we could either put `style={{margin: '0'}}` to all antd code components, (4 or 3 in the issue component), or override it here by using `!important` to specify that we don't want to care for antd component css style definition corresponding to margin attribute of the component.

abishekvashok:
> Let's also address the failing tests :)

Done @r0rshark @gbleaney 

> The filtering has lower priority but if we have time next week we can implement it as well in a separate PR

:in-a-playful-tune: Now I know what to do when I run out of issues 😉

PR #45, opened 2021-08-27
Fix failing ui tests

abishekvashok:
Jest tests were failing because the apollo mock provider was returning
an invalid response (a new attribute was added to issues following PR#43).
Also modifies the snapshot to confront to updates.

ESLint was complaining of an unused import. Readded export option in the filter
which was lost somehow (rebase?)


abishekvashok:
@0xedward thank you for the kind words - learned a lot from you guys this summer :)

PR #49, opened 2021-10-05
makes Web UI display the last character of callable class name for MT issues

esiebomaj:
## Summary

For MT issues, the last character of the callable field is missing
for example, **_y_** is missing in **_MainActivit_** as shown in the image below
This also reflects in the trace frame for the issue

### Before this change
callable in issue
callable in trace

## Test Plan
* Here's a [zip of a SAPP db](https://github.com/MLH-Fellowship/sapp/files/7231200/sapp.db.zip) containing the MT run to help you reproduce the issue, extract it and run the following command to get started looking into this in SAPP:

* Run `python3 -m sapp.cli --database-name sapp.db --debug`

* After the frontend has also started (npm run-script start), go to localhost:3000 to view the changes

### After this change
Notice that the last character of the callable is no longer missing
fixed callable in issue
fixed callable in trace

Relevant issue : [MLH-Fellowship/sapp #3](https://github.com/MLH-Fellowship/sapp/issues/3)

### DIFF
--- a/sapp/ui/frontend/src/HumanReadable.js
+++ b/sapp/ui/frontend/src/HumanReadable.js
@@ -73,7 +73,8 @@ function makeDalvikClassHumanReadable(input: string): string {
   }
 
   const split = input.split('/');
-  return split[split.length - 1].slice(0, -1);
+  const last = split[split.length - 1];
+  return last[last.length-1] === ';' ? last.slice(0, -1) : last;
 }
 
 function makeDalvikHumanReadable(input: string): string {
--- a/sapp/ui/frontend/src/HumanReadable.js
+++ b/sapp/ui/frontend/src/HumanReadable.js
@@ -65,7 +65,7 @@ function makeDalvikParametersHumanReadable(input: string): Array<string> {
   }
 }
 
-function makeDalvikClassHumanReadable(input: string): string {
+export function makeDalvikClassHumanReadable(input: string): string {
   switch (input) {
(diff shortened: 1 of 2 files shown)


sapp/ui/frontend/src/HumanReadable.js
```diff
@@ -73,7 +73,8 @@ function makeDalvikClassHumanReadable(input: string): string {
   }
 
   const split = input.split('/');
-  return split[split.length - 1].slice(0, -1);
+  const last = split[split.length - 1];
```
abishekvashok, line 76: `split.pop()` seems simpler since split isn't used anymore.

abishekvashok (commented):
Tiny nit that others won't possibly care about. Looks good to me :)

sapp/ui/frontend/src/HumanReadable.js
esiebomaj: okay let me use split.pop() instead

esiebomaj: But thinking about this 
what if there is a need to use split later (maybe in the future)
@abishekvashok

abishekvashok: Let's wait for others' opinions. Just my suggestion :)

0xedward:
Thanks for the PR @esiebomaj! :) 

I've imported the PR for internal review and we should have this landed soon

PR #52, opened 2021-10-13
Adds similar issues to sapp

abishekvashok:
Adds similar issues option to sapp to display similar issues associated with each issue.
Adds code in backend to compute similar issues, frontend to display them, and frontend
tests and jest snapshots to honour the new value.

Test plan:
- make sure there's a sapp.db with results injected from pysa or Mariana Trench
- run sapp in debug or production mode:
`python3 -m sapp.cli server --debug` or without `--debug`
- if in production mode, compile the frontend:
```
cd sapp/ui/frontend
npm install
npm run-script build
```
- if in debug mode, start the frontend in development mode:
```
cd sapp/ui/frontend
npm run-script start
```
- goto http://localhost:5000 if in production mode or http://localhost:3000 if in debug mode.
- see similar issues

Fixes: https://github.com/MLH-Fellowship/sapp/issues/9

### DIFF
--- a/sapp/ui/issues.py
+++ b/sapp/ui/issues.py
@@ -5,7 +5,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Dict, FrozenSet, List, NamedTuple, Optional, Set, Union
+from typing import Any, Dict, FrozenSet, List, NamedTuple, Optional, Set, Tuple, Union
 
 import graphene
 from graphql.execution.base import ResolveInfo
@@ -49,6 +49,16 @@ SinkNameText = aliased(SharedText)
 SinkKindText = aliased(SharedText)
 
 
+class SimilarIssueResultType(graphene.ObjectType):
+    issue_id = graphene.ID()
+    score = graphene.Float()
+
+
+class SimilarIssue(NamedTuple):
+    issue_id: DBID
+    score: float
+
+
 # pyre-ignore[13]: unitialized class attribute
 class IssueQueryResultType(graphene.ObjectType):
     concatenated_features: str
@@ -58,6 +68,7 @@ class IssueQueryResultType(graphene.ObjectType):
 
     issue_id = graphene.ID()
     issue_instance_id = graphene.ID()
+    run_id = graphene.ID()
 
     code = graphene.Int()
     message = graphene.String()
@@ -81,6 +92,7 @@ class IssueQueryResultType(graphene.ObjectType):
     min_trace_length_to_sinks = graphene.Int()
 
     warning_message = graphene.String()
+    similar_issues = graphene.List(SimilarIssueResultType)
 
     def resolve_sources(self, info: ResolveInfo) -> List[str]:
         # pyre-ignore[6]: graphene too dynamic.
@@ -109,10 +121,29 @@ class IssueQueryResultType(graphene.ObjectType):
             return warning_message.message
         return ""
 
+    def resolve_similar_issues(self, info: ResolveInfo) -> Set[SimilarIssue]:
+        other_issues = Instance(info.context["session"], DBID(self.run_id)).get()
+
+        for other_issue in other_issues:
+            if (
+                other_issue.issue_instance_id.resolved()
+                # pyre-ignore[16]: grapehene too dynamic
+                == self.issue_instance_id.resolved()
+            ):
+                continue
+            # pyre-ignore[16]: grapehene too dynamic
+            similarity = self.similarity_with(other_issue)
+            if similarity.score > 0.5:
+                # pyre-ignore[16]: grapehene too dynamic
+                self.similar_issues.add(similarity)
+        # pyre-ignore[7]: grapehene too dynamic
+        return self.similar_issues
+
 
 class IssueQueryResult(NamedTuple):
     issue_id: DBID
     issue_instance_id: DBID
+    run_id: DBID
 
     code: int
     message: str
@@ -135,6 +166,8 @@ class IssueQueryResult(NamedTuple):
     source_kinds: FrozenSet[str]
     sink_names: FrozenSet[str]
     sink_kinds: FrozenSet[str]
+
+    similar_issues: Set[SimilarIssue]
 
     @staticmethod
     # pyre-fixme[2]: Parameter annotation cannot be `Any`.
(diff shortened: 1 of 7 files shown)


sapp/ui/schema.py
```diff
@@ -213,6 +213,14 @@ def resolve_issues(
             .where_issue_instance_id_is(issue_instance_id)
             .get()
         )
+
+        other_issues = Instance(session, DBID(run_id)).get()
+        for issue in issues:
+            for other_issue in other_issues:
+                if issue == other_issue:
+                    continue
+                issue.compute_and_update_similarity_with(other_issue)
```
gbleaney, line 222: So this algorithm is exponential in big O notation, which is not going to work for any reasonably sized database. We're going to have to come up with a better method than checking every issue against every other issue. A few potential ideas:

1. If we could come up with some sort of hash that can be used as a similarity measure, this could be reduced to an O(log(n)) lookup in a sorted list
2. If we switched to only showing similar issues on the issue screen, this could be reduced to O(n) per query. This still doesn't make me feel great, but for pretty much every usecase of SAPP it'll probably be OK

gbleaney (changes requested):
As usual, thanks for the PR and for making so many darn contributions to our code :)

sapp/ui/issues.py
```diff
@@ -243,8 +250,27 @@ def __eq__(self, other: object) -> bool:
...
+        score: float = 0.0
+        score += 1 if self.source_names == other.source_names else 0
+        score += 1 if self.sink_names == other.sink_names else 0
+        score += 1 if self.code == other.code else 0
+        score += 1 if self.callable == other.callable else 0
+        score += len(self.sink_kinds.intersection(other.sink_kinds))
+        score += len(self.source_kinds.intersection(other.source_kinds))
+        score = score / (
+            4
+            + len(self.sink_kinds.union(other.sink_kinds))
```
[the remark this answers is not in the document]
abishekvashok: Ahh, yes, I missed the math, thanks for catching this!

sapp/ui/schema.py
abishekvashok: What are your opinions on computing this as a pipeline step and storing it in the db? If we can't do that, I will resort to 2 as I can't come of a hash!

sapp/ui/issues.py
```diff
@@ -126,6 +128,8 @@ class IssueQueryResult(NamedTuple):
     sink_names: FrozenSet[str]
     sink_kinds: FrozenSet[str]
 
+    similar_issues: Set[Tuple[DBID, str]]
```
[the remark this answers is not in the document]
arthaud: Adding on to this: why are we storing a string if it is always computed as a float? We should only convert to string when showing it in the UI.

sapp/ui/schema.py
abishekvashok: @gbleaney @0xedward I came up with a hash, https://www.learndatasci.com/tutorials/building-recommendation-engine-locality-sensitive-hashing-lsh-python/

The query time would be O(1) for each issue (which is something fantastic), but it does have a build up time complexity of O(n). I honestly wanna do this O(n) computation as a pipeline step, store it somewhere, and then just query!

Also for this we may have to add an additional external dependency. But, even if we can't, I can implement this natively.

sapp/ui/issues.py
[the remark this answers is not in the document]
abishekvashok: Okay I will change it to float :)

sapp/ui/issues.py
```diff
@@ -79,6 +79,8 @@ class Meta:
     min_trace_length_to_sources = graphene.Int()
     min_trace_length_to_sinks = graphene.Int()
 
+    similar_issues = graphene.List(graphene.List(graphene.String))
```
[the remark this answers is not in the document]
abishekvashok: @arthaud hmm I wonder why Pyre Github action can't detect this!

abishekvashok:
@arthaud @gbleaney @0xedward done :) Moved it to individual issue's page and fixed the nits

sapp/ui/schema.py
```diff
@@ -218,6 +218,14 @@ def resolve_issues(
             .where_issue_instance_id_is(issue_instance_id)
             .get()
         )
+
+        if len(issues) == 1:
+            other_issues = Instance(session, DBID(run_id)).get()
+            for other_issue in other_issues:
+                if other_issue == issues[0]:
+                    continue
+                issues[0].compute_and_update_similarity_with(other_issue)
```
arthaud, line 227: Can we move this in `compute_and_update_similarity_with`, and make it return the set directly? And I would rename it `similar_issues`.
Most users of `resolve_issues` won't need that, let's not make them pay for it.

arthaud (changes requested):

sapp/ui/schema.py
abishekvashok: @arthaud do you mean to create a seperate endpoint for similar issues?

arthaud: I meant having a separate method on `IssueQueryResultType` that does all this (iterating on all issues and calling `compute_and_update_similarity_with`. There is no reason to have it here.

abishekvashok: Gotcha.

abishekvashok:
@arthaud thank you for the suggestions. Fixed it!

dkgi (changes requested):
Could you make the type checker and tests happy as well? Otherwise this looks reasonable to me :)

abishekvashok:
@dkgi now the type checker and tests are not failing due to errors in this PR. Fixed the tests in #60 . Will fix the type checker in some other PR (as it is unrelated)

abishekvashok:
* ESLint tests are broken.
* Flake8 errors are unrelated


You might need to see how internal pyre-check sees this with those stubs :)

sapp/ui/issues.py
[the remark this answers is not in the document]
abishekvashok: @arthaud sorry for taking so so so long to fix this. I actually couldn't get to work on this until I got to know there were additional stubs used internally which caused this issue. But here is. Hopefully this fixes stuff :)

abishekvashok:
* ESLint is broken In the repository
* Flake8 errors are unrelated
* Pyre check errors are unrelated


You might need to see how internal pyre-check sees this with those stubs :)

PR #60, opened 2021-11-05
Fixes a bug in Runs to return all runs

abishekvashok:
Fixes a bug in runs to return all runs instead of a single one, because
as it seems when we use sqlalchemy.func.count, we need to do a group_by.

Test plan:
- see the tests
- run sapp and goto runs page to see all runs there

Co-authored-by: Esieboma Jeremiah <esiebomaj@gmail.com>


abishekvashok:
tests are failing to things unrelated in this PR. This PR fixes the runs tests :)

PR #62, opened 2021-11-09
Add tests for `warning_messages.py` file

arpancodes:
**Pre-submission checklist**
    - [x] `black .`
    - [x] `usort format .`
    - [x] `flake8`

## Summary

Added tests for the file `/sapp/warning_messages.py`

## Test Plan

Relevant issue: https://github.com/MLH-Fellowship/sapp/issues/11

### DIFF
--- a/sapp/tests/warning_messages_test.py
+++ b/sapp/tests/warning_messages_test.py
@@ -2,15 +2,21 @@ import json
 import pathlib
 from unittest import TestCase
 
+from sqlalchemy import Table
+
 from ..db import DB, DBType
 from ..models import WarningMessage, create as create_model
-from ..warning_messages import update_warning_messages
+from ..warning_messages import update_warning_messages, upsert_entry
 from .fake_object_generator import FakeObjectGenerator
 
 test_metadata = {
     "codes": {
         "1001": "Updated warning message from test",
         "1002": "Updated warning message from test 2",
+    },
+    "codes2": {
+        "2001": "Updated warning message from another test",
+        "2002": "Updated warning message from another test 2",
     },
 }
 
@@ -27,7 +33,7 @@ class WarningMessagesTest(TestCase):
     def tearDown(self) -> None:
         pathlib.Path("sample_metadata.json").unlink()
 
-    def testUpdateWarningMessages(self):
+    def test_update_warning_messages(self):
         with self.db.make_session() as session:
             update_warning_messages(self.db, pathlib.Path("sample_metadata.json"))
             self.assertEqual(
@@ -44,3 +50,35 @@ class WarningMessagesTest(TestCase):
                 .message,
                 "Updated warning message from test 2",
             )
+
+    def test_upsert_entry(self):
+        warning_messages_table = Table(
+            WarningMessage.__tablename__, WarningMessage.metadata
+        )
+        codes = test_metadata["codes2"]
+        with self.db.make_session() as session:
+            with session.connection() as database_connection:
+                for code, message in codes.items():
+                    upsert_entry(
+                        database_connection, warning_messages_table, code, message
+                    )
+                self.assertEqual(
+                    database_connection.execute(
+                        warning_messages_table.select().where(
+                            warning_messages_table.c.code == "2001"
+                        )
+                    )
+                    .first()
+                    .message,
+                    "Updated warning message from another test",
+                )
+                self.assertEqual(
+                    database_connection.execute(
+                        warning_messages_table.select().where(
+                            warning_messages_table.c.code == "2002"
+                        )
+                    )
+                    .first()
+                    .message,
+                    "Updated warning message from another test 2",
+                )
--- a/sapp/tests/warning_messages_test.py
+++ b/sapp/tests/warning_messages_test.py
@@ -1,0 +1,46 @@
+import json
+import pathlib
+from unittest import TestCase
+
+from ..db import DB, DBType
+from ..models import WarningMessage, create as create_model
+from ..warning_messages import update_warning_messages
+from .fake_object_generator import FakeObjectGenerator
+
+test_metadata = {
+    "codes": {
+        "1001": "Updated warning message from test",
+        "1002": "Updated warning message from test 2",
+    },
+}
+
+
+class WarningMessagesTest(TestCase):
+    def setUp(self) -> None:
+        self.db = DB(DBType.MEMORY)
+        create_model(self.db)
+        self.fakes = FakeObjectGenerator()
+
+        with open("sample_metadata.json", "w") as outfile:
+            json.dump(test_metadata, outfile)
+
+    def tearDown(self) -> None:
+        pathlib.Path("sample_metadata.json").unlink()
+
+    def testUpdateWarningMessages(self):
+        with self.db.make_session() as session:
+            update_warning_messages(self.db, pathlib.Path("sample_metadata.json"))
+            self.assertEqual(
+                session.query(WarningMessage)
+                .filter(WarningMessage.code == "1001")
+                .first()
+                .message,
+                "Updated warning message from test",
+            )
+            self.assertEqual(
+                session.query(WarningMessage)
+                .filter(WarningMessage.code == "1002")
+                .first()
+                .message,
+                "Updated warning message from test 2",
+            )


0xedward (changes requested):
Thanks for working on this! :) 

Would you be able to fix the type checking errors? 

```
sapp/tests/warning_messages_test.py:36:4 Missing return annotation [3]: Returning `None` but no return type is specified.
sapp/tests/warning_messages_test.py:40:16 Undefined attribute [16]: Optional type has no attribute `message`.
sapp/tests/warning_messages_test.py:54:4 Missing return annotation [3]: Returning `None` but no return type is specified.
sapp/tests/warning_messages_test.py:63:69 Incompatible parameter type [6]: Expected `int` for 3rd positional only parameter to call `upsert_entry` but got `str`.
```

sapp/tests/warning_messages_test.py
```diff
@@ -0,0 +1,84 @@
...
+}
+
+
+class WarningMessagesTest(TestCase):
+    def setUp(self) -> None:
+        self.db = DB(DBType.MEMORY)
+        create_model(self.db)
+        self.fakes = FakeObjectGenerator()
+
+        with open("sample_metadata.json", "w") as outfile:
```
gbleaney, line 30: The `tempfile` library is slightly more suited for this than rolling your own logid to set up and tear down a test file:
https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile

sapp/tests/warning_messages_test.py
```diff
@@ -0,0 +1,84 @@
...
+        pathlib.Path("sample_metadata.json").unlink()
+
+    def test_update_warning_messages(self) -> None:
+        with self.db.make_session() as session:
+            update_warning_messages(self.db, pathlib.Path("sample_metadata.json"))
+            self.assertEqual(
+                session.query(WarningMessage)
+                .filter(WarningMessage.code == "1001")
+                .first()
+                .__dict__["message"],
```
gbleaney, line 43: Why is this using the `__dict__` method rather then just access the attribute?
```suggestion
                .message,
```

sapp/tests/warning_messages_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_upsert_entry(self) -> None:
+        warning_messages_table = Table(
+            WarningMessage.__tablename__, WarningMessage.metadata
+        )
+        codes = test_metadata["codes2"]
+        with self.db.make_session() as session:
+            with session.connection() as database_connection:
+                for code, message in codes.items():
+                    upsert_entry(
```
gbleaney, line 62: Upsert is a portmanteau of "update" and "insert". The expected behaviour is that it inserts into the DB if the given key is not already present, and updates the row with that key if it is. Right now we're only testing the insert portion. Could we have one case for the update portion?

sapp/tests/warning_messages_test.py
gbleaney: Additionally, this file is currently only used in `test_update_warning_messages`, so there's no need for a dedicated setup/teardown method. The file creation can occur in `test_update_warning_messages` directly

gbleaney (changes requested):
Thanks for adding this test! A few tweaks and then it's ready to land!

sapp/tests/warning_messages_test.py
arpancodes: Actually, it was giving a type error when I was using that (in the previous commit).
```
sapp/tests/warning_messages_test.py:40:16 Undefined attribute [16]: Optional type has no attribute `message`.
```

0xedward: Pyre is probably unhappy because [`sqlalchemy.orm.Query.first()` can return `None`](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.first) if no row is found with the 1001 warning code. 

I think we might have a few options here, but not sure if others have an opinion on the preferred way: 
1. We can write a test case before this one that [checks for the existence of the row](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.exists) with warning code 1001, then change `first()` to [`one_or_none()`](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.one_or_none). That way when this test case runs and 1001 doesn't exist in the table, the previous test case would fail before the exception from `one_or_none()` is thrown.
2. Switch `first()` to `one_or_none()` in this test case and treat the thrown exception as a failed test

arpancodes:
@0xedward and @gbleaney thanks for the review, I have made the requested changes. Please have a look at it again 😄

sapp/tests/warning_messages_test.py
arpancodes: done

sapp/tests/warning_messages_test.py
arpancodes: Did the first one @0xedward

sapp/tests/warning_messages_test.py
arpancodes: added.

sapp/tests/warning_messages_test.py
```diff
@@ -0,0 +1,98 @@
+import json
```
0xedward, line 1: Our linter is bugging us to include the license

```suggestion
# Copyright (c) Facebook, Inc. and its affiliates.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.

import json
```

0xedward (changes requested):
Thanks for making the changes! :) I had one suggestion for managing nested contexts.

Would you also be able to re-run the linters locally and commit those changes?

sapp/tests/warning_messages_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+class WarningMessagesTest(TestCase):
+    def setUp(self) -> None:
+        self.db = DB(DBType.MEMORY)
+        create_model(self.db)
+
+    def test_update_warning_messages(self) -> None:
+        temp = tempfile.NamedTemporaryFile(mode="w+")
+        json.dump(test_metadata, temp)
+        temp.flush()
```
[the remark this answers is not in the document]
arpancodes: added

sapp/tests/warning_messages_test.py
arpancodes: added

sapp/tests/warning_messages_test.py
```diff
@@ -0,0 +1,105 @@
...
+class WarningMessagesTest(TestCase):
+    def setUp(self) -> None:
+        self.db = DB(DBType.MEMORY)
+        create_model(self.db)
+
+    def test_update_warning_messages(self) -> None:
+        with ExitStack() as stack:
+            temp = stack.enter_context(tempfile.NamedTemporaryFile(mode="w+"))
+            json.dump(test_metadata, temp)
+            temp.read()
```
arthaud, line 39: What's the point of the read here, if we are not using the result?

sapp/tests/warning_messages_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class WarningMessagesTest(TestCase):
+    def setUp(self) -> None:
+        self.db = DB(DBType.MEMORY)
+        create_model(self.db)
+
+    def test_update_warning_messages(self) -> None:
+        with ExitStack() as stack:
+            temp = stack.enter_context(tempfile.NamedTemporaryFile(mode="w+"))
```
arthaud, line 37: nitpick: use a better name for this. We try to avoid abbreviations in our codebase.

arthaud (changes requested):

sapp/tests/warning_messages_test.py
arpancodes: If we don't call `.read()`, we are not able to use the file content of it later on the line number 41

sapp/tests/warning_messages_test.py
arpancodes: Will do that.

sapp/tests/warning_messages_test.py
0xedward: It looks like `temp.read()` reads the contents we dumped with `json.dump(test_metadata, temp)`. Without the call, `temp` will be empty when we call `path.Pathlib.read_text(temp.name)`. 

It looks like we might be able to avoid calling `temp.read()` if we set the mode during instantiation of `temp` to `'w+b'`, but then we cannot write a string to `temp`, we would have to encode `test_metadata` as byte object to be able to write it to `temp`. Would you prefer if we approach it this way instead @arthaud?

arthaud: I honestly still don't understand why we would need that `.read()`.
My only idea is that `.read()` flushes some buffer and the next read actually works? What if we use `temp.flush()` instead?
Also, not sure `w+b` will change anything.

arpancodes: `temp.flush()` works as well, I can change it to that 😄

PR #63, opened 2021-11-19
Added tests for filters.py in sapp/ui

esiebomaj:
**Pre-submission checklist**
    - [x] `black .`
    - [x] `usort format .`
    - [x] `flake8`

## Summary
The PR adds tests for sapp/frontend/ui/filters.py and increases the coverage for that file from 0% to 80%

## Test Plan
to run the tests in test_run.py 
`python -m coverage run -m unittest sapp.ui.tests.filter_test && python -m coverage report --show-missing --skip-covered --skip-empty`

```
Name                        Stmts   Miss  Cover   Missing
---------------------------------------------------------
sapp/ui/filters.py          137    47    80%   
```

Relevant Issue : [mlh-fellowship/sapp #11](https://github.com/MLH-Fellowship/sapp/issues/11)


esiebomaj:
while trying to fix some typing issues in the tests I noticed that the `delete_filters` function in [`sapp\ui\filters.py`](https://github.com/facebook/sapp/blob/main/sapp/ui/filters.py) takes an argument `filter_names` of type `Tuple[str]`. This implies that it should receive a tuple with only one string in it. 

But the name delete_filters implies that it should be able to accept more than one filter name. The logic in the function is also for deleting more than 1 filters
Moreover, there is already a delete_filter function that deletes just one filter

So in this commit (01c3f4d) I changed the type annotation for filter_names argument from Tuple[str] with Tuple[str, ...]
to reflect that the argument takes multiple filter_names

PR #65, opened 2021-12-02
Fix line-number off by one error in MT parser

abishekvashok:
Fix line-number being off by one in MT parser.
MT for some reason, reports line number as 0 which can lead to issues in
the frontend. Add one to it to prevent this problem.

This can also be something that is wrong with Mariana Trench as the
general convention is to report line numbers from 1 rather than 0.
Hence this commit is pretty open to debate.

Fixes: https://github.com/MLH-Fellowship/sapp/issues/4

### DIFF
--- a/sapp/pipeline/mariana_trench_parser.py
+++ b/sapp/pipeline/mariana_trench_parser.py
@@ -133,6 +133,8 @@ class Position(NamedTuple):
     def from_json(position: Dict[str, Any], method: Method) -> "Position":
         path = position.get("path", UNKNOWN_PATH)
         line = position.get("line", UNKNOWN_LINE)
+        if line != UNKNOWN_LINE:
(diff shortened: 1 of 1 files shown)


abishekvashok:
As mentioned in the commit message, I would love to have views on this small patch.

arthaud:
I'm pretty puzzled why we would need this for the SAPP UI but we don't need this for our internal UI. Especially since we don't have the same problem for Pysa. Something is wrong here, we should find out what it is.

abishekvashok:
@arthaud I am also puzzled why MT reports line numbers from 0. The UI does a `line - 1` to convert the line number to array index and hence the crash. (It happens only when we try to mark something on line 1 though)

Also, I was asked to move this to the frontend. Hence, closing this :)

PR #66, opened 2021-12-12
Match GraphQL CORS header with frontend hostname and port in debug mode

EdOverflow:
**Pre-submission checklist**
    - [x] `black .`
    - [x] `usort format .`
    - [x] `flake8`

## Summary

`start.js` allows users to specify the host and port to serve the frontend from by relying on the `HOST` and `PORT` environment variables. If these environment variables are not set, the frontend resorts to a default host `0.0.0.0` and port `3000`.

```js
const DEFAULT_PORT = parseInt(process.env.PORT, 10) || 3000;
const HOST = process.env.HOST || '0.0.0.0';
```

The way the frontend extracts the data is by querying the `/graphql` endpoint which sets a CORS header in debug mode via `server.py`.

A bug was discovered when users set the `HOST` and/or `PORT` environment variable to something other than `0.0.0.0` and `3000` respectively. The CORS header for the `/graphql` endpoint would prevent the frontend from being able to query it.

This Pull Request introduces a solution by matching the code for specifying the frontend host and port in the CORS header set by `server.py`.

```diff
-        CORS(application, resources={r"/graphql": {"origins": "http://localhost:3000"}})
+        default_port = int(os.environ.get("PORT"), 10) or 3000
+        host = os.environ.get("HOST") or "localhost"
+        CORS(
+            application,
+            resources={
+                r"/graphql": dict(origins="http://" + host + ":" + str(default_port))
+            },
+        )
```

My only concern with this current implementation is that the behaviour of `parseInt()` in JavaScript may differ from Python's `int()` function for specific edge cases and may result in unexpected issues.

Additionally, the `README.md` file was updated to include a warning concerning the reliance on the two aforementioned environment variables.

## Test Plan

The following steps should allow one to reproduce the behaviour described above in the `main` branch.

```
$ export PORT=8080
$ python3 -m sapp.cli server --debug
$ cd sapp/ui/frontend && npm run-script start
```

image

The changes introduced by this PR should resolve the test case above and no CORS errors should be encountered.


sapp/ui/server.py
```diff
@@ -99,5 +99,12 @@ def start_server(
     if static_resources:
         application.static_folder = static_resources
     if debug:
-        CORS(application, resources={r"/graphql": {"origins": "http://localhost:3000"}})
+        default_port = int(os.environ.get("PORT"), 10) or 3000
+        host = os.environ.get("HOST") or "localhost"
+        CORS(
+            application,
+            resources={
+                r"/graphql": dict(origins="http://" + host + ":" + str(default_port))
```
0xedward, line 107: I think we prefers f-strings or `str.format` over concatenation

```suggestion
                r"/graphql": dict(origins="http://{hostname}:{port}".format(hostname=host, port=default_port)
```

sapp/ui/server.py
```diff
@@ -99,5 +99,12 @@ def start_server(
     if static_resources:
         application.static_folder = static_resources
     if debug:
-        CORS(application, resources={r"/graphql": {"origins": "http://localhost:3000"}})
+        default_port = int(os.environ.get("PORT"), 10) or 3000
```
0xedward, line 102: > My only concern with this current implementation is that the behaviour of parseInt() in JavaScript may differ from Python's int() function for specific edge cases and may result in unexpected issues.

Like you mentioned, `int(os.environ.get("PORT"), 10)` here will result in a `TypeError`. The suggestion below should fix that:
```suggestion
        default_port = os.environ.get("PORT") or 3000
```

0xedward (changes requested):
Hey @EdOverflow, thanks for fixing this and writing up such a detailed PR! :)

I'll import this PR for internal review. Sorry again about the delay in getting back to you :(

README.md
```diff
@@ -295,14 +295,14 @@ Installing dependencies for frontend:
...
+To run SAPP with hot reloading of the Web UI, you need have the frontend and backend running simultaneously. In a production environment, the frontend application is compiled and served directly by the backend exposed on port `13337`. But in a development environment, the frontend runs in port `3000` by default if the `PORT` environment variable is not set and the backend runs in port `13337`. You can indicate to SAPP to run in the development environment with the `debug` flag.
 
 Run the flask server and react app in development mode:
 ```shell
 (sapp) $ python3 -m sapp.cli server --debug
 (sapp) $ cd sapp/ui/frontend && npm run-script start
 ```
-Then visit `http://localhost:3000`
+Then visit `http://localhost:3000` (or `http://localhost:<PORT>` if you have set the `PORT` environment variable).
```
0xedward, line 305: Would we also be able to add something in the docs to cover how setting the `HOST` env var will change what URL to visit?

```python
        host = os.environ.get("HOST") or "localhost"
```

EdOverflow: Thank you for taking the time to review this Pull Request, @0xedward. Nice to bump into you again. :)

I agree with this suggestion. I will add some documentation concerning the `HOST` environment variable.

sapp/ui/server.py
```diff
@@ -99,5 +99,12 @@ def start_server(
     if static_resources:
         application.static_folder = static_resources
     if debug:
-        CORS(application, resources={r"/graphql": {"origins": "http://localhost:3000"}})
+        default_port = os.environ.get("PORT") or 3000
+        host = os.environ.get("HOST") or "localhost"
+        CORS(
+            application,
+            resources={
+                r"/graphql": dict(origins="http://{hostname}:{port}".format(hostname=host, port=default_port)
```
0xedward, line 107:
```suggestion
                r"/graphql": dict(origins="http://{hostname}:{port}".format(hostname=host, port=default_port))
```

Looks like we are missing a closing parentheses here

https://github.com/facebook/sapp/runs/4885612036?check_suite_focus=true

0xedward (changes requested):
Thanks for making the changes! There's a small thing that needs to fixed and we should be able to merge this shortly afterwards :)

sapp/ui/server.py
```diff
@@ -99,5 +99,12 @@ def start_server(
...
         application.static_folder = static_resources
     if debug:
-        CORS(application, resources={r"/graphql": {"origins": "http://localhost:3000"}})
+        default_port = os.environ.get("PORT") or 3000
+        host = os.environ.get("HOST") or "localhost"
+        CORS(
+            application,
+            resources={
+                r"/graphql": dict(origins="http://{hostname}:{port}".format(hostname=host, port=default_port))
+            },
```
0xedward, line 108: Our linter triggered on this line - https://github.com/adamchainz/flake8-comprehensions#c408-unnecessary-dictlisttuple-call---rewrite-as-a-literal

```suggestion
            resources={
                r"/graphql": {"origins": "http://{hostname}:{port}".format(hostname=host, port=default_port)}
            },
```

0xedward (changes requested):
We're so close to merging this! :)

Would you mind running the linters on this PR to fix the linter warnings?

- https://github.com/facebook/sapp/runs/4944490776?check_suite_focus=true
- https://github.com/facebook/sapp/runs/4944490454?check_suite_focus=true

EdOverflow:
The aforementioned linting warnings should be resolved now, @0xedward.

PR #90, opened 2023-03-03
Quick fix for #89: Bump pyre-extensions version in requirements.txt

### LINKED ISSUE #89: Error installing `fb-sapp` due to dependency issue
Hi Folks,

While trying to install `fb-sapp` along with `pyre-check`, I am receiving the following error:
```
Because no versions of pyre-check match >0.9.18,<0.10.0
 and pyre-check (0.9.18) depends on pyre-extensions (>=0.0.29), pyre-check (>=0.9.18,<0.10.0) requires pyre-extensions (>=0.0.29).
And because fb-sapp (0.5.4) depends on pyre-extensions (0.0.27)
 and no versions of fb-sapp match >0.5.4,<0.6.0, pyre-check (>=0.9.18,<0.10.0) is incompatible with fb-sapp (>=0.5.4,<0.6.0).
So, because test-project depends on both fb-sapp (^0.5.4) and pyre-check (^0.9.18), version solving failed.
```

The `pyre-check` version I am trying to install is the latest as of writing this issue, which is `0.9.18` and I have tried poetry and pip both. This installation works correctly with pyre-check version `0.9.15`.

I believe the `pyre-extension` package is outdated in fb-sapp (which is 0.0.27) and needs to be in-line with the latest `pyre-check` version (which is 0.0.30)

abishekvashok:
Bump pyre-extensions version in requirements.txt
Since, these are extensions are for the typing module, I think we can remove the version information as well.

**Pre-submission checklist**
    - [x] `black .`
    - [x] `usort format .`
    - [x] `flake8`

## Summary

A simple version bump for pyre-extensions.
In order to prevent dependency issues in sapp, we decided to pin versions in the requirements.txt. Now a package that pyre-check introduces a dependency conflict as it depends on a newer version of a package (pyre-extensions)
Bumping pyre-extensions which extend the typing module. Since it is non-breaking module (pyre-extensions), we could potentially unpin the versioning information and allow pyre-check to dictate the version. However, I could be wrong here.

## Test Plan

> Ran 102 tests in 5.983s

The frontend tests are failing however, but I strongly believe them to be un-related to this PR.

Fixes #89

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ ipython<8.4.0
 munch==2.5.0
 prompt-toolkit==3.0.29
 Pygments==2.12.0
-pyre-extensions==0.0.27
+pyre-extensions==0.0.30
 SQLAlchemy<1.4
 traitlets==5.2.2.post1
 typing-extensions==4.2.0


requirements.txt
```diff
@@ -9,7 +9,7 @@ ipython<8.4.0
 munch==2.5.0
 prompt-toolkit==3.0.29
 Pygments==2.12.0
-pyre-extensions==0.0.27
+pyre-extensions==0.0.30
```
stroxler, line 12: Thanks for the fix, looks good to me!

I'll try to get this landed today

stroxler (approved):

abishekvashok:
@stroxler maybe also consider:
we could potentially unpin the versioning information for pyre-extensions and allow `pyre-check` to dictate the version. However, I might be wrong here.

Cc: @0xedward who pinned versions :)

PR #93, opened 2023-07-14
Adds SARIF traces support to SAPP

the-storm:
**Pre-submission checklist**
- [✅] I've ran the following linters locally and fixed lint errors related to the files I modified in this PR
    - [✅] `black .`
    - [✅] `usort format .`
    - [✅] `flake8`
- [✅] I've installed dev dependencies `pip install -r requirements-dev.txt` and completed the following:
    - [✅] I've ran tests with `./scripts/run-tests.sh` and made sure all tests are passing

## Summary
This PR adds a sarif output to SAPP traces. Currently SARIF output from SAPP doesn't output the trace of an issue. It just outputs the root node making it triage almost impossible. 

Example of SARIF output on [exercise_3](https://github.com/facebook/pyre-check/blob/main/documentation/pysa_tutorial/exercise3/views.py) from Pysa tutorial is

```
{
          "ruleId": "5001",
          "level": "warning",
          "message": {
            "text": "Data from [UserControlled] source(s) may reach [RemoteCodeExecution] sink(s)"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "views.py"
                },
                "region": {
                  "startLine": 24,
                  "startColumn": 19,
                  "endColumn": 36
                }
              }
            }
          ]
        }
```
Notice locations section only shows the root of where the issue is happening but not a full trace
Below is the SARIF output of the same issue after applying this PR
```
{
          "ruleId": "5001",
          "level": "warning",
          "message": {
            "text": "Data from [UserControlled] source(s) may reach [RemoteCodeExecution] sink(s)"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "views.py"
                },
                "region": {
                  "startLine": 24,
                  "startColumn": 19,
                  "endColumn": 36
                }
              }
            }
          ],
          "codeFlows": [
            {
              "threadFlows": [
                {
                  "locations": [
                    {
                      "location": {
                        "physicalLocation": {
                          "artifactLocation": {
                            "uri": "views.py",
                            "uriBaseId": "%SRCROOT%"
                          },
                          "region": {
                            "startLine": 14,
                            "startColumn": 16,
                            "endColumn": 28
                          }
                        },
                        "message": {
                          "text": "flow from views.get_operator_safe(...result...) -[into]-> Obj{django.http.request.HttpRequest.POST}(...source...)  features: ['has:first-index', 'first-index:operator']"
                        }
                      },
                      "nestingLevel": 0
                    },
                    {
                      "location": {
                        "physicalLocation": {
                          "artifactLocation": {
                            "uri": "views.py",
                            "uriBaseId": "%SRCROOT%"
                          },
                          "region": {
                            "startLine": 22,
                            "startColumn": 16,
                            "endColumn": 42
                          }
                        },
                        "message": {
                          "text": "flow from views.operate_on_twos(...root...) -[into]-> views.get_operator_safe(...result...) via 1 propagators features: ['always-via:obscure:model', 'always-via:format-string', 'always-via:tito']"
                        }
                      },
                      "nestingLevel": 1
                    },
                    {
                      "location": {
                        "physicalLocation": {
                          "artifactLocation": {
                            "uri": "views.py",
                            "uriBaseId": "%SRCROOT%"
                          },
                          "region": {
                            "startLine": 24,
                            "startColumn": 19,
                            "endColumn": 36
                          }
                        },
                        "message": {
                          "text": "flow from views.operate_on_twos(...root...) -[into]-> eval(...sink...)"
                        }
                      },
                      "nestingLevel": 2
                    }
                  ]
                }
              ]
            }
          ]
        }
```
Notice the new object in SARIF output `codeFlows` 
The PR is heavily inspired by [`_generate_trace_from_issue`](https://github.com/facebook/sapp/blob/ab2aa2e586878e251806c32c13ede53aa36ee91f/sapp/ui/interactive.py#L857-L900) from the interactive output for SAPP

## Test Plan
Analyzed exercise 3 with Pysa https://github.com/facebook/pyre-check/blob/main/documentation/pysa_tutorial/exercise3/ and attached the SARIF before and after
SARIF before this PR: [exercise3_old.sarif.txt](https://github.com/facebook/sapp/files/12049937/exercise3_old.sarif.txt)
SARIF after this PR: [exercise3.sarif.txt](https://github.com/facebook/sapp/files/12049936/exercise3.sarif.txt)
Screenshot before this PR Screenshot 2023-07-14 at 12 33 46

Screenshot after this PR Screenshot 2023-07-14 at 12 33 25

PS: I noticed there are no unit tests for SARIF output. I am happy to follow-up later with A PR that add some unit tests for SARIF export. 
PS: I have also run the pre-submission checklist but I have noticed that the tests are already broken without any changes from my side.

### DIFF
--- a/sapp/sarif.py
+++ b/sapp/sarif.py
@@ -51,7 +51,8 @@ class SARIF:
             "pysa": (5000, 6000),
         }
         driver_json = {}
-        if tool == "pysa":
+        self.tool = tool
+        if self.tool == "pysa":
             driver_json["name"] = "Pysa"
             driver_json["informationUri"] = "https://github.com/facebook/pyre-check/"
 
@@ -64,11 +65,26 @@ class SARIF:
             for rule in tool_warning_messages:
                 rules_json.append({"id": str(rule.code), "name": rule.message})
             driver_json["rules"] = rules_json
+        elif self.tool == "mariana-trench":
+            driver_json["name"] = "Mariana Trench"
+            driver_json[
+                "informationUri"
+            ] = "https://github.com/facebook/mariana-trench/"
+
+            tool_warning_messages = get_warning_message_range(
+                session,
+                self._tool_warning_code_ranges[self.tool][0],
+                self._tool_warning_code_ranges[self.tool][1],
+            )
+            rules_json = []
+            for rule in tool_warning_messages:
+                rules_json.append({"id": str(rule.code), "name": rule.message})
+            driver_json["rules"] = rules_json
         else:
             raise NotImplementedError
 
         self.driver = driver_json
-        self.results = [issue.to_sarif() for issue in filtered_issues]
+        self.results = [issue.to_sarif(session, self.tool) for issue in filtered_issues]
 
     def to_json(self, indent: int = 2) -> str:
         return json.dumps(self, cls=SARIFEncoder, indent=indent)
--- a/sapp/sarif.py
+++ b/sapp/sarif.py
@@ -51,40 +51,28 @@ class SARIF:
             "pysa": (5000, 6000),
         }
         driver_json = {}
-        self.tool = tool
-        if self.tool == "pysa":
+        if tool == "pysa":
             driver_json["name"] = "Pysa"
             driver_json["informationUri"] = "https://github.com/facebook/pyre-check/"
-
-            tool_warning_messages = get_warning_message_range(
-                session,
-                self._tool_warning_code_ranges[tool][0],
-                self._tool_warning_code_ranges[tool][1],
-            )
-            rules_json = []
-            for rule in tool_warning_messages:
-                rules_json.append({"id": str(rule.code), "name": rule.message})
-            driver_json["rules"] = rules_json
-        elif self.tool == "mariana-trench":
+        elif tool == "mariana-trench":
             driver_json["name"] = "Mariana Trench"
             driver_json[
                 "informationUri"
             ] = "https://github.com/facebook/mariana-trench/"
-
-            tool_warning_messages = get_warning_message_range(
-                session,
-                self._tool_warning_code_ranges[self.tool][0],
-                self._tool_warning_code_ranges[self.tool][1],
-            )
-            rules_json = []
-            for rule in tool_warning_messages:
-                rules_json.append({"id": str(rule.code), "name": rule.message})
-            driver_json["rules"] = rules_json
         else:
             raise NotImplementedError
 
+        tool_warning_messages = get_warning_message_range(
+            session,
+            self._tool_warning_code_ranges[tool][0],
+            self._tool_warning_code_ranges[tool][1],
+        )
+        rules_json = []
+        for rule in tool_warning_messages:
+            rules_json.append({"id": str(rule.code), "name": rule.message})
+        driver_json["rules"] = rules_json
         self.driver = driver_json
-        self.results = [issue.to_sarif(session, self.tool) for issue in filtered_issues]
+        self.results = [issue.to_sarif(session, tool) for issue in filtered_issues]
 
     def to_json(self, indent: int = 2) -> str:
         return json.dumps(self, cls=SARIFEncoder, indent=indent)
--- a/sapp/ui/trace.py
+++ b/sapp/ui/trace.py
@@ -34,7 +34,6 @@ from ..models import (
 from ..sarif_types import (
     SARIFCodeflowLocationObject,
     SARIFCodeflowsObject,
-    SARIFPhyicalLocationObject,
 )
 from . import run
 from .issues import IssueQueryResult
@@ -121,20 +120,22 @@ class TraceFrameQueryResult(NamedTuple):
         return self.callee_port in LEAF_NAMES
 
     def get_clean_caller(self, tool: str) -> str:
-        return self._clean_callable_name(self.caller, tool)
+        return self.get_human_readable_caller(self.caller, tool)
 
     def get_clean_callee(self, tool: str) -> str:
-        return self._clean_callable_name(self.callee, tool)
+        return self.get_human_readable_caller(self.callee, tool)
 
-    def _clean_callable_name(self, name: str, tool: str, depth: int = 1) -> str:
+    def get_human_readable_caller(self, name: str, tool: str, depth: int = 1) -> str:
         if name in LEAF_NAMES:
             return name
         if tool == "mariana-trench":
-            package_and_class, method_name = name.split(".")
-            method_name = method_name.split(":")[0]  # parse
-            package_and_class = package_and_class.strip(";").split("/")[
-                depth * -1
-            ]  # parse Lpackage1/package2/class;.
+            # convert name from dalvik byte code e.g., Lpackage1/package2/class;.Func1 to human readable name class.Func1
+            package_class_and_method_name = name.split(".")
+            if len(package_class_and_method_name) != 2:
+                return name
+            package_and_class, method_name = package_class_and_method_name
+            method_name = method_name.split(":")[0]
+            package_and_class = package_and_class.strip(";").split("/")[depth * -1]
             return f"{package_and_class}.{method_name}"
         else:
             return name
@@ -468,16 +469,15 @@ def to_sarif(
     trace_tuples = _create_trace_tuples(
         reversed(postcondition_navigation)
     ) + _create_trace_tuples(precondition_navigation)
-    codeflows = [{"threadFlows": [{"locations": []}]}]
+    codeflows = []
     nesting_level = 0
     for t in trace_tuples:
         location = _sarif_codeflow_location_from_trace_tuple(
-            t.trace_frame, tool, nesting_level, True
+            t.trace_frame, tool, nesting_level, output_features=output_features
         )
-        codeflows[0]["threadFlows"][0]["locations"].append(location)
+        codeflows.append(location)
         nesting_level += 1
-
-    return codeflows
+    return [{"threadFlows": [{"locations": codeflows}]}]
 
 
 def _create_trace_tuples(
@@ -513,7 +513,7 @@ def _sarif_codeflow_location_from_trace_tuple(
     trace_region = {}
     if trace_frame.callee_location:
         trace_region = trace_frame.callee_location.to_sarif()
-    location = {
+    return {
         "location": {
             "physicalLocation": {
                 "artifactLocation": {
@@ -528,4 +528,3 @@ def _sarif_codeflow_location_from_trace_tuple(
         },
         "nestingLevel": nesting_level,
     }
-    return location
--- a/sapp/ui/trace.py
+++ b/sapp/ui/trace.py
@@ -3,7 +3,18 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Dict, List, NamedTuple, Optional, Sequence, Set, Tuple, Union
+from typing import (
+    Any,
+    Dict,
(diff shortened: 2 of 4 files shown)


sapp/sarif.py
```diff
@@ -64,11 +65,26 @@ def __init__(
...
+
+            tool_warning_messages = get_warning_message_range(
+                session,
+                self._tool_warning_code_ranges[self.tool][0],
+                self._tool_warning_code_ranges[self.tool][1],
+            )
+            rules_json = []
+            for rule in tool_warning_messages:
+                rules_json.append({"id": str(rule.code), "name": rule.message})
+            driver_json["rules"] = rules_json
```
arthaud, line 82: Instead of copy/pasting this part from lines 59-67, could we move it outside of the if/elif?

sapp/ui/trace.py
```diff
@@ -403,3 +438,94 @@ def trace_kind_to_shared_text_kind(trace_kind: Optional[TraceKind]) -> SharedTex
...
+                },
+                "region": trace_region,
+            },
+            "message": {
+                "text": f"flow from {trace_frame.get_clean_caller(tool)}(...{trace_frame.caller_port}...) -[into]-> {trace_frame.get_clean_callee(tool)}(...{trace_frame.callee_port}...) {titos} {features_str}".strip()
+            },
+        },
+        "nestingLevel": nesting_level,
+    }
+    return location
```
arthaud, line 531: Could we move all the logic regarding sarif into `sarif.py`?

sapp/ui/trace.py
```diff
@@ -103,6 +120,25 @@ def from_record(
     def is_leaf(self) -> bool:
         return self.callee_port in LEAF_NAMES
 
+    def get_clean_caller(self, tool: str) -> str:
+        return self._clean_callable_name(self.caller, tool)
+
+    def get_clean_callee(self, tool: str) -> str:
+        return self._clean_callable_name(self.callee, tool)
+
+    def _clean_callable_name(self, name: str, tool: str, depth: int = 1) -> str:
```
arthaud, line 129: nitpick: Internally we call this "human readable name", could we use the same here?

Also: what does the `depth` parameter do? Maybe remove it since it's never used.

sapp/ui/trace.py
```diff
@@ -103,6 +120,25 @@ def from_record(
...
+        if name in LEAF_NAMES:
+            return name
+        if tool == "mariana-trench":
+            package_and_class, method_name = name.split(".")
+            method_name = method_name.split(":")[0]  # parse
```
arthaud, line 134: nit: comment doesn't really help
```suggestion
            method_name = method_name.split(":")[0]
```

sapp/ui/trace.py
```diff
@@ -103,6 +120,25 @@ def from_record(
...
+            package_and_class = package_and_class.strip(";").split("/")[
+                depth * -1
+            ]  # parse Lpackage1/package2/class;.
```
arthaud, line 137: ditto, comment doesn't really help. Maybe a useful comment would be an example of the full name at the top of this code.

sapp/ui/trace.py
```diff
@@ -403,3 +438,94 @@ def trace_kind_to_shared_text_kind(trace_kind: Optional[TraceKind]) -> SharedTex
...
+        postcondition_initial_frames,
+        set(issue.source_kinds),
+        set(issue.sink_kinds),
+    )
+    precondition_navigation = navigate_trace_frames(
+        session,
+        precondition_initial_frames,
+        set(issue.source_kinds),
+        set(issue.sink_kinds),
+    )
```
arthaud, line 467: I quickly checked those functions and I think it might not support transforms, maybe you want to double check if that's important to you

sapp/ui/trace.py
```diff
@@ -403,3 +438,94 @@ def trace_kind_to_shared_text_kind(trace_kind: Optional[TraceKind]) -> SharedTex
...
+    codeflows = [{"threadFlows": [{"locations": []}]}]
+    nesting_level = 0
+    for t in trace_tuples:
+        location = _sarif_codeflow_location_from_trace_tuple(
+            t.trace_frame, tool, nesting_level, True
+        )
+        codeflows[0]["threadFlows"][0]["locations"].append(location)
+        nesting_level += 1
+
+    return codeflows
```
arthaud, line 480: Instead of creating `codeflows` to this whole structure, just use it as a list and create the structure last:

```suggestion
    codeflows = []
    nesting_level = 0
    for t in trace_tuples:
        location = _sarif_codeflow_location_from_trace_tuple(
            t.trace_frame, tool, nesting_level, True
        )
        codeflows.append(location)
        nesting_level += 1

    return [{"threadFlows": [{"locations": codeflows}]}]
```

sapp/ui/trace.py
```diff
@@ -403,3 +438,94 @@ def trace_kind_to_shared_text_kind(trace_kind: Optional[TraceKind]) -> SharedTex
...
+        set(issue.sink_kinds),
+    )
+    trace_tuples = _create_trace_tuples(
+        reversed(postcondition_navigation)
+    ) + _create_trace_tuples(precondition_navigation)
+    codeflows = [{"threadFlows": [{"locations": []}]}]
+    nesting_level = 0
+    for t in trace_tuples:
+        location = _sarif_codeflow_location_from_trace_tuple(
+            t.trace_frame, tool, nesting_level, True
```
arthaud, line 475: I would add the name of the boolean parameter for clarity

```suggestion
            t.trace_frame, tool, nesting_level, output_features=True
```

sapp/ui/trace.py
```diff
@@ -403,3 +438,94 @@ def trace_kind_to_shared_text_kind(trace_kind: Optional[TraceKind]) -> SharedTex
...
+    if output_features:
+        frame_features = [
+            text.contents
+            for text in trace_frame.shared_texts
+            if text.kind is SharedTextKind.FEATURE
+        ]
+        if frame_features:
+            features_str = f"features: {frame_features}"
+        if trace_frame.titos and len(trace_frame.titos.split(";")) > 0:
+            titos = f"via {len(trace_frame.titos.split(';'))} propagators"
```
arthaud, line 512: This is pre-existing but it looks like we store a list as a string separated by `;`. We should really just have the list within `trace_frame`.

Also I'm not sure what those `titos` are compared to `features`.
Finally, "propagators" sounds weird to me but no strong feelings :p

sapp/ui/trace.py
```diff
@@ -403,3 +438,94 @@ def trace_kind_to_shared_text_kind(trace_kind: Optional[TraceKind]) -> SharedTex
...
+    trace_region = {}
+    if trace_frame.callee_location:
+        trace_region = trace_frame.callee_location.to_sarif()
+    location = {
```
arthaud, line 516: nit: 
```suggestion
    return {
```

arthaud (changes requested):
Thanks for the work!

sapp/ui/trace.py
the-storm: I was thinking to use the depth to split not he package_and_class name for something like this 


`A.B.C.D.E.class.function_name` so the depth 0 can only include `class.function_name` and as you increase the depth you get more packages in the name. I know in Java things can become ugly so that's why I thought about the depth. Forgot to use it though 😅

the-storm: > nitpick: Internally we call this "human readable name", could we use the same here?

Gotcha!

sapp/ui/trace.py
the-storm: Yep I realised that too. Not for now I am using transforms only for sanitizer purposes but soon we will need for valid flows so I will probably follow-up here. I will need also to update the interactive UI to be consistent

sapp/ui/trace.py
the-storm: > Finally, "propagators" sounds weird to me but no strong feelings :p

happy if you have better naming (remember I am terrible with names :D). I will as ChatGTP for help :D

sapp/sarif.py
```diff
@@ -51,7 +51,8 @@ def __init__(
             "pysa": (5000, 6000),
         }
         driver_json = {}
-        if tool == "pysa":
+        self.tool = tool
```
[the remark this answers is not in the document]
the-storm: hmm I was thinking we will need tool to add relevant information to the SARIF later tool-specific and I felt it'd be helpful to know which tool we are creating the SARIF for, but yes I don't think we are using it now so I will remove it and once I have a use-case I can bring it back

sapp/ui/trace.py
the-storm: > Also I'm not sure what those titos are compared to features.

I am not comparing I am just trying to give a summary for each frame through the number of titos and list of features available in the frame.

sapp/ui/trace.py
arthaud: Well IMO anything related to sarif should go into `sarif.py`. If we start putting function from every format into `trace.py` it will get overwhelming soon.
`trace.py` should provide a clean API that allows any formatter to get the data and do it's formatting.
Now I understand this is a pre-existing problem, so we can probably try to fix that later.

sapp/ui/trace.py
```diff
@@ -103,6 +119,27 @@ def from_record(
     def is_leaf(self) -> bool:
         return self.callee_port in LEAF_NAMES
 
+    def get_clean_caller(self, tool: str) -> str:
+        return self._human_readable_name(self.caller, tool)
+
+    def get_clean_callee(self, tool: str) -> str:
+        return self._human_readable_name(self.callee, tool)
```
arthaud, line 126: Can you also call these `get_human_readable_caller`, etc?

arthaud (changes requested):

arthaud:
I still believe we should move that code into `sarif.py`.

the-storm:
> I still believe we should move that code into sarif.py.



let's merge this and I follow-up with something that re-organises SARIF relevant code into sarif.py

arthaud (approved):

arthaud (approved):

arthaud:
FYI There are type errors we need to fix:
[tools/sapp/sapp/sarif_types.py:45:0] Invalid type parameters [24]: Generic type `list` expects 1 type parameter, use `typing.List[<element type>]` to avoid runtime subscripting errors.
[tools/sapp/sapp/ui/issues.py:257:36] Incompatible parameter type [6]: In call `dict.__setitem__`, for 2nd positional argument, expected `Union[Dict[str, str], List[Dict[str, Dict[str, Union[Dict[str, int], Dict[str, str]]]]], str]` but got `List[Dict[str, List[Dict[str, List[typing.Any]]]]]`.
[tools/sapp/sapp/ui/trace.py:514:4] Incompatible return type [7]: Expected `Dict[str, Union[Dict[str, Union[Dict[str, Union[Dict[str, int], Dict[str, str]]], Dict[str, str]]], Dict[str, str]]]` but got `Dict[str, Union[Dict[str, Union[Dict[str, Dict[typing.Any, typing.Any]], Dict[str, str]]], int]]`.
Type `Dict[str, Union[Dict[str, Union[Dict[str, Union[Dict[str, int], Dict[str, str]]], Dict[str, str]]], Dict[str, str]]]` expected on line 534, specified on line 494. See [https://pyre-check.org/docs/errors...](https://pyre-check.org/docs/errors#covariance-and-contravariance) for mutable container errors.

arthaud:
I fixed the type errors and I am landing this. Please pull before working on follow-ups since I had to make many changes.

PR #96, opened 2023-07-25
refactor sarif messages

the-storm:
**Pre-submission checklist**
- [✅] I've ran the following linters locally and fixed lint errors related to the files I modified in this PR
    - [✅] `black .`
    - [✅] `usort format .`
    - [✅] `flake8`
- [✅]] I've installed dev dependencies `pip install -r requirements-dev.txt` and completed the following:
    - [✅] I've ran tests with `./scripts/run-tests.sh` and made sure all tests are passing
    - Tests are failing but already failing without this PR 

## Summary
This PR refactors the sarif relevant functions in the codebase and move it into `sarif.py` (centralized location). This was discussed as a follow-up to https://github.com/facebook/sapp/pull/93. 
## Test Plan
Run sapp on [exercise_3](https://github.com/facebook/pyre-check/blob/main/documentation/pysa_tutorial/exercise3/views.py) from Pysa tutorial and compared the output with this PR and without this PR and noticed no changes in the output.  SARIF output of [exercise3.txt](https://github.com/facebook/sapp/files/12164331/exercise3.txt)

### DIFF
--- a/sapp/sarif.py
+++ b/sapp/sarif.py
@@ -6,16 +6,28 @@
 # pyre-strict
 
 import json
-from dataclasses import dataclass, field
 from json import JSONEncoder
 from typing import Dict, List, Set, Tuple, Union
+
+from sapp.models import SharedTextKind
+from sapp.pipeline import SourceLocation
 
 from sqlalchemy.orm import Session
 from typing_extensions import TypeAlias
 
 from .queries import get_warning_message_range
-from .sarif_types import SARIFResult
+from .sarif_types import (
+    SARIFCodeflowLocationInnerObject,
+    SARIFCodeflowLocationObject,
+    SARIFCodeflowsObject,
+    SARIFRegionObject,
+    SARIFResult,
+    SARIFSeverityLevel,
+    SARIFThreadFlowObject,
+)
+from .ui import trace
 from .ui.issues import IssueQueryResult
+from .ui.trace import TraceFrameQueryResult, TraceKind
 
 
 SARIFOutput: TypeAlias = Dict[
@@ -35,13 +47,9 @@ SARIFOutput: TypeAlias = Dict[
 ]
 
 
-@dataclass
 class SARIF:
     version: str = "2.1.0"
     schema: str = "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json"  # noqa
-    _tool_warning_code_ranges: Dict[str, Tuple[int, int]] = field(default_factory=dict)
-    driver: Dict[str, Union[str, List[Dict[str, str]]]] = field(default_factory=dict)
-    results: List[SARIFResult] = field(default_factory=list)
 
     def __init__(
         self, tool: str, session: Session, filtered_issues: Set[IssueQueryResult]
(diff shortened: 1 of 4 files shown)


sapp/sarif.py
```diff
@@ -42,6 +55,7 @@ class SARIF:
     _tool_warning_code_ranges: Dict[str, Tuple[int, int]] = field(default_factory=dict)
     driver: Dict[str, Union[str, List[Dict[str, str]]]] = field(default_factory=dict)
     results: List[SARIFResult] = field(default_factory=list)
+    tool: str = "pysa"
```
the-storm, line 58: moved the tool to a field since it is used internally and not worth it passing it around different functions

sapp/sarif.py
```diff
@@ -42,6 +55,7 @@ class SARIF:
     _tool_warning_code_ranges: Dict[str, Tuple[int, int]] = field(default_factory=dict)
```
the-storm, line 55: I don't this should be a `dataclass` anymore /cc @arthaud  wdyt?

arthaud: right, a `dataclass` with a "private" mutable member is odd. Maybe we could just make this a normal class.
Since it's pre-existing I guess it's fine to keep or fix in a follow-up.

sapp/sarif.py
```diff
@@ -64,19 +79,137 @@ def __init__(
...
+            precondition_initial_frames,
+            set(issue.source_kinds),
+            set(issue.sink_kinds),
+        )
+        trace_tuples = trace.create_trace_tuples(
+            reversed(postcondition_navigation)
+        ) + trace.create_trace_tuples(precondition_navigation)
+        codeflows: List[SARIFCodeflowLocationObject] = []
+        nesting_level = 0
+        for t in trace_tuples:
```
arthaud, line 150: nitpick: avoid abbreviations (`t`)

sapp/sarif.py
```diff
@@ -64,19 +79,137 @@ def __init__(
...
         for rule in tool_warning_messages:
             rules_json.append({"id": str(rule.code), "name": rule.message})
         driver_json["rules"] = rules_json
         self.driver = driver_json
-        self.results = [issue.to_sarif(session, tool) for issue in filtered_issues]
+        self.results = [
+            self.issue_to_sarif(session, issue) for issue in filtered_issues
+        ]
+
+    def issue_to_sarif(
```
arthaud, line 94: nitpick: should all these method start with `_` because they are not supposed to be called from outside the class?

arthaud (approved):
LGTM, Thanks

sapp/sarif.py
the-storm: I am not sure. I think they may be called from outside e.g., imagine a case where some UI that builds on top of SARIF and then it can call this class to get different parts of the SARIF, but I am flexible if we want to keep this private i.e add `_`.

sapp/sarif.py
```diff
@@ -35,13 +47,13 @@
...
-@dataclass
 class SARIF:
     version: str = "2.1.0"
     schema: str = "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json"  # noqa
-    _tool_warning_code_ranges: Dict[str, Tuple[int, int]] = field(default_factory=dict)
-    driver: Dict[str, Union[str, List[Dict[str, str]]]] = field(default_factory=dict)
-    results: List[SARIFResult] = field(default_factory=list)
+    _tool_warning_code_ranges: Dict[str, Tuple[int, int]] = dict()
+    driver: Dict[str, Union[str, List[Dict[str, str]]]] = dict()
+    results: List[SARIFResult] = list()
```
arthaud, line 55: The problem is that those fields are initialized only once at the class level. This will lead to strange bugs:

```
class Foo:
  results = list()

Foo().results.append(1)
print(Foo().results) # shows [1] even though it's a different instance.
```

If we are not using `@dataclass` anymore we need to initialize all fields manually in the constructor.

arthaud (changes requested):

sapp/sarif.py
the-storm: oh yes 🤦‍♂️

